PR #39, opened 2014-01-24
Feature: Option for BULK MODE SCAN from NATIVE APPLICATION is added

mharisce:
# Feature: 

Option for BULK MODE SCAN from NATIVE APPLICATION is added. Check ZXing Test application for sample use. See button "Bulk Scan Anything".
## Minor Note:

Many files have very simple changes due to removal of java 1.7 features like <>, switch on string etc. It is because i am using eclipse and by default Android doesn't support java 1.6

Signed-off-by: mharisce m.haris.ce@gmail.com

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,12 @@ target
 private
 .idea
 *.iml
+/core-lib
+/integration-lib
+
+/android/.classpath
+/android/.project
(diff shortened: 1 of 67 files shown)


kjkrum:
Is there any way to enable this feature in an `Intent` and receive multiple scans in `onActivityResult(...)`?

PR #267, opened 2014-12-12
Featurehistory

Bfinity:
I made additions to the res/values/strings file adding two strings for the history preference xml. 
I made additions to the res/xml/pref file adding a checkbox preference for the turning off history.
I made addition to the PreferenceActivity file adding a constant string variable for the history.
I made addition to the CaptureActivity file adding an if statement to check history preference before storing history after livescan. 

I made these additions using a text editor and also have the patch saved if emailing a patch would be better.

### DIFF
--- a/android/res/values/strings.xml
+++ b/android/res/values/strings.xml
@@ -104,6 +104,8 @@
   <string name="preferences_disable_barcode_scene_mode_title">No barcode scene mode</string>
   <string name="preferences_disable_continuous_focus_summary">Use only standard focus mode</string>
   <string name="preferences_disable_continuous_focus_title">No continuous focus</string>
+  <string name="preferences_disable_history">Barcode history is not automatically saved</string>
+  <string name="preferences_disable_history_title">Turn off barcode history</string>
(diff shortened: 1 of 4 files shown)


srowen:
Feel free to reopen if you want to fix the issues above in this PR.

android/res/values/strings.xml
```diff
@@ -104,6 +104,8 @@
   <string name="preferences_disable_barcode_scene_mode_title">No barcode scene mode</string>
   <string name="preferences_disable_continuous_focus_summary">Use only standard focus mode</string>
   <string name="preferences_disable_continuous_focus_title">No continuous focus</string>
+  <string name="preferences_disable_history">Barcode history is not automatically saved</string>
```
Bfinity: Just this single string or both strings that reference history? And, where can I find the tool?

srowen: All the new strings. Hm, I remembered though that you have to have an API key for Google Translate. It only costs cents but probably not worth you bothering. Unless you add them by hand I can just run the tool later.

PR #269, opened 2014-12-17
return full text contain \r

khajer:
some barcode contain '\r' ,so i think return full text not replace "\r" -> ""
thx


srowen:
Reopen if there is a response to my question here. The PR fails unit tests anyway.

PR #272, opened 2014-12-17
minSize from provided sizes

andrejb-dev:
create minSize Dimension from provided width and height. If minsize is provided inside hints, it still gets overwritten
it was not tested, tests failed at aztec barcode


srowen:
I'm seeing the same problem as reported in https://github.com/zxing/zxing/issues/361 and I think it's due to this change. Here the min / max size are really setting the min / max _symbol_ dimension, not the rendered image size. This causes it to fail for example if you ask for a 300x300 image as there is no DM symbol that large.

This reveals another lesser problem, which is probably the original issue here, that the DM encoder doesn't try to then render an image of the given dimensions. You get a tiny image whose dimensions match the symbol's. That's better than not working but still not exactly consistent.

Zavael:
ok, after so much time passed i dont dig deeply again to this, but the behaviour of providing the preferred dimensions should be unified across different Writer implementations. Feel free to close this if its not.

PR #279, opened 2015-01-02
Fix: Android Studio - switch to the if / else if

gustavohma:
The switch was replaced by the if / else if when we use the R class, because an error occurs in Android Studio.


srowen:
No, the code is correct as is. You are trying to use this code as a library, but it is not supposed to be used that way .

gustavohma:
Okay. Good luck.

PR #280, opened 2015-01-04
dependency-info.html: sbt should use %

nafg:
%% is for libraries cross-built against different scala versions.

### DIFF
--- a/dependency-info.html
+++ b/dependency-info.html
@@ -201,7 +201,7 @@
 <div class="section">
 <h3>SBT<a name="SBT"></a></h3><a name="SBT"></a>
 <div class="source">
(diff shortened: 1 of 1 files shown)


srowen:
This is auto generated by the maven site plugin. It would be best to propose changes to this output in the plugin project.

nafg:
Can you give me a more targeted pointer?

srowen:
This is the site plugin: http://maven.apache.org/plugins/maven-site-plugin/
Does the formulation not work at all? or just suboptimal for projects that don't publish Scala-specific "_2.10" artifacts?
I assume it would be simple to just expand the boilerplate text it generates to cover this case.

nafg:
Apparently I'm not the first to notice: http://jira.codehaus.org/browse/MPIR-296
Seems it's fixed for maven-project-info-reports-plugin 2.8 which seems to not be out yet.

Thanks!

PR #283, opened 2015-01-08
Removing unnecessary code from CameraManager

mertemin:
This patch simply updates CameraManager and OpenCameraInterface by removing unneeded code.


srowen:
That's fine. I'll probably follow on to make a constant for the -1 value and add some javadoc

PR #292, opened 2015-01-19
Revert "Fix PDF417 numeric encoding grouping index check"

Tommygun78:


srowen:
Why do you think this should be reverted?

PR #296, opened 2015-01-23
Added the ability to disable history saving

MicheleMas:
Added a new preference to enable and disable history saving (enabled by default).

### DIFF
--- a/android/res/values-it/strings.xml
+++ b/android/res/values-it/strings.xml
@@ -120,7 +120,7 @@
   <string name="preferences_remember_duplicates_summary">Mantieni scansioni multiple del codice a barre nella stessa Cronologia</string>
   <string name="preferences_remember_duplicates_title">Ricorda duplicati</string>
   <string name="preferences_history_summary">Mantieni le tue scansioni nella Cronologia</string>
-  <string name="preferences_history_title">Aggiungi alla cronologia</string>
+  <string name="preferences_history_title">Aggiungi alla Cronologia</string>
   <string name="preferences_result_title">Impostazioni risultato</string>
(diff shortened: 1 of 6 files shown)


srowen:
Yes, I think this is OK. There's already a means to disable saving by `Intent`, so OK to have a preference. I think that `HistoryManager` could check this preference and remember it at construction time, and then use it in both `addHistoryItem` and `addHistoryItemDetails`. That way the logic is more centralized. Are you up for making that change?

MicheleMas:
It would be better in this way, but I think that it could have some update issues. From what I can see in debug mode, the `HistoryManager` instance used for saving the records is not updated when I close the preferences activity, so it would keep the old pref value until a restart.
If you want I can save a `SharedPreferences` object at construction time in `HistoryManager`, and check the boolean value when the two methods (`addHistoryItem` and `addHistoryItemDetails`) are called. In this way it should be logically correct, and there should not be update issues.

You know the code way better than me obviously, so if I'm wrong just tell me, and I will implement it as you suggested before.

srowen:
A-ha, right. The initialization of `HistoryManager` could simply be moved to `onResume`. It is not at all expensive to recreate it. Then it would be initialized every time the main activity resumed.

MicheleMas:
Ok done. Now `HistoryManager` is initialized inside `onResume`, and the boolean value is loaded at construction time.
Just one thing, `addHistoryItem` is now skipped if the pref is disabled, but I didn't modified `addHistoryItemDetails` as pointed out in the comment that i found inside the method.

```// As we're going to do an update only we don't need need to worry about the preferences; if the item wasn't saved it won't be udpated```

PR #308, opened 2015-02-12
A small correction on Spanish translation

mhcuervo:
According to the DRAE (Royal Spanish Academy), there's no noun derived from "escáner" (scanner) that represents the act of scanning something. So, taking into account the original sentence in English: "No barcode scans have been recorded", I think a more appropriate translation would be "Ninguna lectura de código de barras se ha registrado aún", which uses the word "lectura" (read - as the act of reading something) instead.
Please, also notice a fix on agreement.


srowen:
Thank you!

mhcuervo:
You're welcome! I think the earliest commit in this pull request can be omitted. Sorry, I'm not very familiar with Github. The problem seems to be related to the fact that I've pulled your latest changes to my local repository directly, not through Github.

srowen:
Yeah I could have squashed it but I was too lazy.

Issue #314, opened 2015-02-19
Make IntentIntegrator REQUEST_CODE parametrizable


eMaringolo:
The current implementation of ```initiateScan()``` uses a fixed request code when calling ```startActivityForResult(...)```, which could be bypassed by calling ```createIntent()``` and starting the intent activity externally. 

But ```parseActivityResult()``` cannot be bypassed, because REQUEST_CODE is hardwired in the code.

It would be desirable to have a version of ```initiateScan()``` and ```parseActivityResult()``` that receives the request code as argument.

srowen:
I think that's fine, do you want to make a PR that adds a getter/setter for a request code, which defaults to the current value?

renatosc:
I am in need of this. Already added a getter/setter. Can I make the PR?

PR #316, opened 2015-02-20
Fixes Github Issue #314

eMaringolo:
Instead of using the request code as a parameter, preserves the previous api of ```parseActivityResult()``` and delegates its implementation to a new ```parseActivityResult()``` which adds a boolean parameter to determine whether to use the static (hardcoded) REQUEST_CODE, if such parameter is true, it behaves as before, if it false it doesn't check against the request code.

A better alternative would be to honor original ```parseActivityResult```'s ```requestCode``` in the code, but it would require a change in the API.

### DIFF
--- a/android-integration/src/main/java/com/google/zxing/integration/android/IntentIntegrator.java
+++ b/android-integration/src/main/java/com/google/zxing/integration/android/IntentIntegrator.java
@@ -106,6 +106,7 @@ import android.util.Log;
  * @author Isaac Potoczny-Jones
  * @author Brad Drehmer
  * @author gcstang
+ * @author Esteban Maringolo
  */
 public class IntentIntegrator {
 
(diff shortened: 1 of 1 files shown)


srowen:
This doesn't check any request code though. I thought you needed to be able to configure what code you look for? You need to check it at some level.

eMaringolo:
You specify the request code in the ```startActivityForResult(..)``` of the scan intent caller, once it returns via ```onActivityResult(...)``` you determine whether to match it or not. It is the responsibility of the caller to match the requestCode.

So if I want to use a custom request code I could do:
```java
        IntentIntegrator integrator = new IntentIntegrator(this);
        integrator.setDesiredBarcodeFormats(IntentIntegrator.ONE_D_CODE_TYPES);
        integrator.setResultDisplayDuration(0);
        startActivityForResult(integrator.createScanIntent(), CUSTOM_REQUEST_CODE1);
```

And then:

So if I want to use a custom request code I could do:
```java
protected void onActivityResult(int requestCode, int resultCode, Intent data) {
       if (requestCode == CUSTOM_REQUEST_CODE1) {
             IntentResult result = IntentIntegrator.parseActivityResult(requestCode, resultCode, data, false);
             /* ...  request code 1 ... */
       } else if (requestCode == CUSTOM_REQUEST_CODE2) {
           /* ...  request code 2 ... */
       }
}
```

In this case the request codes are static, but they could be dynamic.
I hope this is self explanatory.

Thank you.

srowen:
Yes, but I thought you wanted to expose control over the request code to the caller. Right now you can't; I thought this would entail another overloaded `initiateScan` method. I don't recognize the `createScanIntent()` method.

Why make the caller handle this logic? the point of the integration code is to handle it for you. At least it does at the moment. Yes it's getting a little overkill to pass in the request code and expected one just for it to check equality, but I think it's less convoluted than checking it and also passing in a new flag that disables a similar check.

eMaringolo:
You need the caller to handle the logic because there is only ```onActivityResult()``` handler per caller, you can't define a result listener for the Integrator. So one way or another, you'll need to map the requestCode (wrapped by the integrator or not) to the caller, to know which integrator made which request.

Regarding the ```createScanIntent()``` I just realized it's not there, I thought it was part of the official ZXing project, but it turns out to be part of https://github.com/journeyapps/zxing-android-embedded but not of its upstream (maybe it was added later). So basically I modified the IntentIntegrator API of a version I'm not consuming. :no_mouth: 

We can leave this PR on hold until I clarify this.

srowen:
Yes, so the caller has to specify the request code at the outset. I think that can be tacked on to the API without trouble. Yes there is no `createScanIntent` here otherwise.

PR #323, opened 2015-03-04
Support for Data Matrix Rectangular Extensions

dehnhard:
This is a functional (but not clean) implementation of Data Matrix Rectangular Extensions as defined in the specification http://www.aim-d.de/images/stories/pdfs/Arbeitskreise/Barcode/aim_spec_dmre_2014-09-22.pdf

The implementation is divided in 3 separate commits:

https://github.com/zxing/zxing/commit/c815a4babb3708f1ae3e1115bb5dbb61d50cd4b8:
This adds the rectangles to the datamatrix definition tables.

https://github.com/zxing/zxing/commit/f2a32a45f053215175fa6750a006106b44a0705a:
This changes the threshold for determining a rectangular datamatrix form. This is necessary, because the old threshold excluded some of the new rectangles. It has the side effect of test failures in the datamatrix code, but AFAICS it has no effect on the functionality itself.

https://github.com/zxing/zxing/commit/e6713ae554c44fb158e49e7139a5441165db71af:
Some of the new forms resulted in an ArrayOutOfBoundsException. As I am not able to fix this cleanly I implemented this worlaround. The Exception may probably be fixed easily by a developer with more in-depth knowledge...

### DIFF
--- a/core/src/main/java/com/google/zxing/common/BitMatrix.java
+++ b/core/src/main/java/com/google/zxing/common/BitMatrix.java
@@ -131,7 +131,10 @@ public final class BitMatrix implements Cloneable {
    */
   public boolean get(int x, int y) {
     int offset = y * rowSize + (x / 32);
-    return ((bits[offset] >>> (x & 0x1f)) & 1) != 0;
+    if(offset<bits.length)
+      return ((bits[offset] >>> (x & 0x1f)) & 1) != 0;
+    else
+      return false;
   }
 
   /**
@@ -142,12 +145,12 @@ public final class BitMatrix implements Cloneable {
    */
   public void set(int x, int y) {
     int offset = y * rowSize + (x / 32);
-    bits[offset] |= 1 << (x & 0x1f);
+    if(offset<bits.length) bits[offset] |= 1 << (x & 0x1f);
   }
 
   public void unset(int x, int y) {
     int offset = y * rowSize + (x / 32);
-    bits[offset] &= ~(1 << (x & 0x1f));
+    if(offset<bits.length) bits[offset] &= ~(1 << (x & 0x1f));
   }
 
   /**
@@ -158,7 +161,7 @@ public final class BitMatrix implements Cloneable {
    */
   public void flip(int x, int y) {
     int offset = y * rowSize + (x / 32);
-    bits[offset] ^= 1 << (x & 0x1f);
+    if(offset<bits.length) bits[offset] ^= 1 << (x & 0x1f);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


wfred101:
The extended rectangular data matrix symbols can be created by using a fork of the zint library 
http://sourceforge.net/u/oehhar/zint/ci/master/tree/

The extension seems to be adopted to ISO
http://www.iso.org/iso/iso_technical_committee.html?commid=45332

6versus9:
First time

oehhar:
Great work ! DMRE is really important and will have lots of use cases.

For printing, one may use zint:
Binaries:
http://www.eurodatacouncil.org/de/doks-downloads.html
Source Code:
https://sourceforge.net/projects/zint/

Thank you, great work,
Harald

srowen:
I committed a variation on this in https://github.com/zxing/zxing/commit/395dc6b83f4793a73d2fe69c42a7c9ee6b1273ee

dehnhard:
Glad to see this old work is getting attention.

@srowen Looks like you left out f2a32a4. IIRC this means that some DMRE codes will work, but those who exceed the original height width ratio will fail.
When I did this work I dove just enough into the code to incorporate DMRE and haven't been sure about possible side effects (on the original datamatrix codes). As of now I think it's safe to say that this patches have been used and proven stable in public apps for many years, including this commit.

srowen:
@dehnhard this code doesn't seem to exist anymore in the master branch. That's why I didn't merge that change. Do you have an updated change?

dehnhard:
It looks like there has been a rewrite of the Datamatrix Detector in 2018, which moved that code to line 74:

https://github.com/zxing/zxing/commit/c13216ce08889bdec3a158aa3161db77dda5ad54#diff-96280025a27511b1486b1d5c83e6a43785b896ef1599552c26a228a59e06aaaeR74

Can you work with that or shall I provide a new PR?

srowen:
I don't see how to translate that directly now. Yes please open a PR if you can.

dehnhard:
I put the equivalent change in #1350

A word of warning though: The above mentioned public use of the DMRE patch extends to ZXing version 3.3.3 (because it's the last version supporting Java 8) and therefor doesn't include this "Detector rewrite". Also there is a suspicious comment a line below talking about squares instead of rectangles. On the other hand if the rectangular dimensions of original DataMatrix are still working, there is no reason for the DMRE extensions to fail.

To get at the bottom of this would mean to analyse the algorithm, manually test codes and optimally writing unit tests, but this exceeds what I can do in my free time. I can just state that if I happen to work on this in the future again I will gladly provide the results here for public use, if possible.

PR #343, opened 2015-03-20
Fixed the cameraId changing on powerbutton press

nathbenjwolf:
When using the QR code reader with a front facing camera (non-default camera) if you turn the screen off and on (pressing power button) the camera being used changes to the default camera. This occurs because when the screen is turned off and on the surfaceView is not destroyed causing a different  behavior in CaptureActivity.onResume(). The surfaceView not being destroyed causes the CameraManager to open the camera before setting the cameraId. This pull request has reordered these operations to make sure the manual setting of cameraId inside CaptureActivity.onResume() occurs before the CameraManager opening the camera.

For more information see the following post:
https://groups.google.com/forum/#!topic/zxing/CguxsAm0SjE

### DIFF
--- a/android/src/com/google/zxing/client/android/CaptureActivity.java
+++ b/android/src/com/google/zxing/client/android/CaptureActivity.java
@@ -176,16 +176,6 @@ public final class CaptureActivity extends Activity implements SurfaceHolder.Cal
 
     resetStatusView();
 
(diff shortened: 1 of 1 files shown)


srowen:
Yes that seems fine to just move the initialization later. Thanks!

PR #347, opened 2015-03-31
Support android.support.v4.app.Fragment in IntentIntegrator (fixes #291)

### LINKED ISSUE #291: using initiateScan() in android.support.v4.app.Fragment is not invoking onActivityResult()
I tried the following inside the android.support.v4.app.Fragment:
```java
 IntentIntegrator integrator = new IntentIntegrator(getActivity());
```
The constructor for the Fragment is only available for the native Fragment provided by the SDK in the device not the support-v4 or package.
After deeper investigation i find out that the REQUEST_CODE:
https://github.com/zxing/zxing/blob/6c522f127a0ac40b2e14518528c9547018e044b2/android-integration/src/main/java/com/google/zxing/integration/android/IntentIntegrator.java#L112

value seems the issue. It prevents later in the FragmentActivity here:
http://grepcode.com/file/repository.grepcode.com/java/ext/com.google.android/android/5.0.0_r1/android/support/v4/app/FragmentActivity.java/#150
that the index is not 0. Then it traverses through all fragments and invokes onActivityResult() properly.
I set a breakpoint with the debugger and modified the REQUEST_CODE and it works. I haven't tested it if this implies other problems.

Would be nice to have your feedback. Thank you very much for reading.

Zillode:
This PR supports IntentIntegrator() for ```android.support.v4.app.Fragment```

*I license the work to the project under the project's open source license.*

### DIFF
--- a/android-integration/pom.xml
+++ b/android-integration/pom.xml
@@ -26,6 +26,11 @@
       <groupId>com.google.android</groupId>
       <artifactId>android</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.android</groupId>
+      <artifactId>support-v4</artifactId>
+      <version>r7</version>
+    </dependency>
   </dependencies>
 
   <parent>
--- a/android-integration/src/main/java/com/google/zxing/integration/android/IntentIntegrator.java
+++ b/android-integration/src/main/java/com/google/zxing/integration/android/IntentIntegrator.java
@@ -141,6 +141,7 @@ public class IntentIntegrator {
   
   private final Activity activity;
   private final Fragment fragment;
+  private final android.support.v4.app.Fragment fragmentV4;
 
   private String title;
   private String message;
@@ -155,6 +156,7 @@ public class IntentIntegrator {
   public IntentIntegrator(Activity activity) {
     this.activity = activity;
     this.fragment = null;
+    this.fragmentV4 = null;
     initializeConfiguration();
   }
 
@@ -166,6 +168,20 @@ public class IntentIntegrator {
   public IntentIntegrator(Fragment fragment) {
     this.activity = fragment.getActivity();
     this.fragment = fragment;
+    this.fragmentV4 = null;
+    initializeConfiguration();
+  }
+
+  /**
+   * @param fragment {@link android.support.v4.app.Fragment} invoking the integration.
+   *  {@link #startActivityForResult(Intent, int)} will be called on the
+   *  {@link android.support.v4.app.Fragment} instead of an {@link Activity} or
+   *  {@link Fragment}.
+   */
+  public IntentIntegrator(android.support.v4.app.Fragment fragment) {
+    this.activity = fragment.getActivity();
+    this.fragment = null;
+    this.fragmentV4 = fragment;
     initializeConfiguration();
   }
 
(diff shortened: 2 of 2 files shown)


srowen:
Hm, I think this is something that was removed a long time ago from this class. How much need is there today to support fragments across older versions of Android? I think the logic in the past was that there's not much need for this anymore.

Zillode:
Well there is at least @drindt, @jiggo and us :) I guess many other developers copy/paste the code or use a workaround. We have a bit more than 130 active users that require the support-v4 branch of our rather small app. Your call

ProdigiumKami:
I am having trouble getting zxing to respond to my intent to scan a barcode, I am using your app to scan a barcode as described in you API but the app doesnt open :/ it insted crashes my app. After multiple tries, i decided to proceed with my development efforts by adding a catch block which sends an Intent to the store to download the app. The play store responds to the Intent, so I'm wondering what's wrong here. ive tried a bunch of stuff for months.

M13Kd:
#m13kd @m13kd #LALILULELO ¤MSFT LIVE Rocks 
XBOX COMMUNITY AMBASSADORS
Https://Xcl.Xbox.Com/

alexis18:
:)

A024806:
鄭文仲

2015/7/6 下午12:33於 alexis18 <notifications@github.com>寫道：

Nemisis:
Maybe I am wrong but I am sure that most people use the Support Libraries implementation of Fragment over the standard built in Fragment. That's because, if you want to use the AppCompat library (to support material design across all platforms), then you will need to use the Support implementation.

drindt:
@Nemisis it seems that no one cares about this.

Nemisis:
That's fine. It's easy enough to include in your project and make the switch yourself.

drindt:
@Nemisis you can have a look at my quick-fix [here](https://github.com/zxing/zxing/issues/291).

axd1967:
add me to the list of people that need this merge...

pacien:
Could this be reconsidered for merging? `android.app.Fragment` is now deprecated.

> This class was deprecated in API level 28.
> Use the Support Library Fragment for consistent behavior across all devices and access to Lifecycle. 

https://developer.android.com/reference/android/app/Fragment

srowen:
This change supports really old versions of Android, right? I don't think support for API 4 helps API 28? This app is basically EOL and will not target past API 19 as it would need a rewrite of the permissions handling.

pacien:
> I don't think support for API 4 helps API 28?

It actually does now. The new Android Architecture components are only supporting `android.support.v4.app.Fragment` which implements the new [`LifecycleOwner`](https://developer.android.com/reference/android/arch/lifecycle/LifecycleOwner) interface. That support library now fully replaces the stock `android.app.Fragment`.

> This app is basically EOL and will not target past API 19 as it would need a rewrite of the permissions handling.

Would it require a lot of changes? I depend on this application in a couple of projects. I might be able to spare some time and submit a PR for this.

srowen:
I see. If it's still working with API 28 and probably releases for a while, though deprecated, I think I'd leave it. Nothing is broken per se, as I understand?

In November the app won't be even uploadable to the app store anymore, so any new releases would necessarily be a bigger rewrite anyway.

PR #370, opened 2015-05-04
Translation updated to Portuguese-BR

BrunoMioto:


srowen:
This is purely for my own interest, but are these edits just fixing Google Translate or do they maybe represent differences between Portuguese in Brazil and Portugal?

BrunoMioto:
Yeah, I think this Google Translation's translation used to be to Portuguese-BR. My translation is for Portuguese-BR. Maybe you should create two docs for this different translations.

PR #393, opened 2015-05-20
Use result metadata to return Code39 Full-ASCII decoded value and checksum status

l-dobrev:
The proposed change will:
* attempt to decode Code39 in Full-ASCII form, and if successful will return it as a result meta-data.
* attempt to calculate the checksum and return the state as a result meta-data
* display the Code39 Full-ASCII result in the Android scanner screen

TODO:
* Handle codes that have a checksum and are encoded using the Code39 Full-ASCII form

### DIFF
--- a/core/src/main/java/com/google/zxing/ResultMetadataType.java
+++ b/core/src/main/java/com/google/zxing/ResultMetadataType.java
@@ -93,5 +93,15 @@ public enum ResultMetadataType {
    * parity is given with it.
    */
   STRUCTURED_APPEND_PARITY,
-  
+
+  /**
+   * For Code39 this contains the decoded contents with the Full-ASCII range.
+   */
+  CODE_39_EXTENDED,
+
+  /**
+   * For Code39 this contains a boolean marking whether the checksum is valid.
+   */
+  CODE_39_CHECKSUM_CORRECT,
+
 }
--- a/core/src/main/java/com/google/zxing/oned/Code39Reader.java
+++ b/core/src/main/java/com/google/zxing/oned/Code39Reader.java
@@ -22,6 +22,7 @@ import com.google.zxing.DecodeHintType;
 import com.google.zxing.FormatException;
 import com.google.zxing.NotFoundException;
 import com.google.zxing.Result;
+import com.google.zxing.ResultMetadataType;
 import com.google.zxing.ResultPoint;
 import com.google.zxing.common.BitArray;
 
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/google/zxing/oned/Code39Reader.java
```diff
@@ -140,16 +141,25 @@ public Result decodeRow(int rowNumber, BitArray row, Map<DecodeHintType,?> hints
...
-        total += ALPHABET_STRING.indexOf(decodeRowResult.charAt(i));
-      }
-      if (result.charAt(max) != ALPHABET[total % 43]) {
+    // Calculate the checksum. Validate it if it was requested.
+    int max = result.length() - 1;
+    int total = 0;
+    for (int i = 0; i < max; i++) {
+      total += ALPHABET_STRING.indexOf(decodeRowResult.charAt(i));
+    }
+    Boolean checksumFlag;
```
srowen: Is the rest of this section simpler as...

```
boolean checksumPassed = result.charAt(max) != ALPHABET[total % 43];
if (usingCheckDigit) {
  if (checksumPassed) {
    result.setLength(max);
  } else {
    throw ChecksumException.getChecksumInstance();
  }
}
```

core/src/main/java/com/google/zxing/ResultMetadataType.java
```diff
@@ -93,5 +93,15 @@
    * parity is given with it.
    */
   STRUCTURED_APPEND_PARITY,
-  
+
+  /**
+   * For Code39 this contains the decoded contents with the Full-ASCII range.
+   */
+  CODE_39_EXTENDED,
```
srowen: Call this "CODE_39_FULL_ASCII" ? IIRC there is another format mode called 'extended' but I can't find it now.

core/src/main/java/com/google/zxing/ResultMetadataType.java
```diff
@@ -93,5 +93,15 @@
...
+
+  /**
+   * For Code39 this contains a boolean marking whether the checksum is valid.
+   */
+  CODE_39_CHECKSUM_CORRECT,
```
srowen: I might be dreaming too far here, but since there are a few formats like this, what about `OPTIONAL_CHECKSUM_CORRECT` or something? no point promising it for required checksums but this might have a similar role for other formats.

core/src/main/java/com/google/zxing/ResultMetadataType.java
srowen: Ah yeah, just found that Code 93 also has this idea, so may not even be right to restrict this to Code 39 in the name: http://en.wikipedia.org/wiki/Code_93

l-dobrev:
To be rewritten

Issue #394, opened 2015-05-20
Multibyte characters are not handled properly in PDF417 encoder 


strguntbr:
Trying to encode the string "1%§s ?aG$" with PDF417 encoder (Charset UTF-8, Compaction Byte) throw a "Non-encodable character detected" exception.

The problem seems to be that § is encoded as "c2 a7" in unicode and thus for all characters after the § the index in the input-string does not match the index in the encoded byte array.

srowen:
Hm yeah, I don't know a lot about this code path, but I can imagine the problem and your fix looks reasonable. Let me think about it a little more, but since tests pass and it doesn't seem to affect other behavior, can't hurt much to merge I think.

PR #395, opened 2015-05-20
Multibyte fix

strguntbr:
Unittest and possible fix for https://github.com/zxing/zxing/issues/394


strguntbr:
Unfortunately i seemed to introduce a new bug in this fix. I called String.substring(start, length) instead of String.substring(start, end). I'll fix it in the next days and create another pull request that hopefully does not contain any more bugs :)

PR #397, opened 2015-05-23
Decode HTML entities in titles

kmark:
This allows page titles with HTML entities like `&amp;` to display properly. Before, no decoding would be performed and `&amp;` would be shown where an ampersand should be. Additionally, decoding entities prevents the following string length check from overcounting.

To replicate this bug scan an image with a URL pointing to http://stackoverflow.com/questions/tagged/android The title displayed within the app will be something like: `Frequent &#39;android&#39; Questions - Stack Overflow` After this commit, the app will display the correct decoded title: `Frequent 'android' Questions - Stack Overflow`

Alternatively, scan this to replicate:
http://stackoverflow.com/questions/tagged/android

### DIFF
--- a/android/src/com/google/zxing/client/android/result/supplement/TitleRetriever.java
+++ b/android/src/com/google/zxing/client/android/result/supplement/TitleRetriever.java
@@ -16,6 +16,7 @@
 
 package com.google.zxing.client.android.result.supplement;
 
+import android.text.Html;
 import android.widget.TextView;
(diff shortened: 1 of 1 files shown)


srowen:
Good point, that part isn't necessarily supposed to have HTML but can contain escapes. This seems like an easy way to parse that.

lstella114:
; )

kmark:
Thanks for the quick merge

PR #404, opened 2015-06-30
Code93Writer class

codewell4:
The quick and dirty CODE_93 writer. Made with many ugly hacks but it should work as per specification

Didn't reuse some of the Code93Reader constants and generally ignored other classes coding styles

Had to redesign appendPattern function from OneDimensionalCodeWriter.java because it didn't output my codes right - it is included below the class

Didn't test it in the core since I only have Android development environment set. But it works in Android project so it should generally work in Java :)


srowen:
OK, that's fine but if it's a quick hack without tests I'm not going to merge it. It's here for anyone to look at I suppose.

PR #407, opened 2015-07-01
Naturalize's suggestions

condda:
Naturalize's renaming suggestions to improve the codebase consistency

### DIFF
--- a/core/src/main/java/com/google/zxing/multi/GenericMultipleBarcodeReader.java
+++ b/core/src/main/java/com/google/zxing/multi/GenericMultipleBarcodeReader.java
@@ -86,8 +86,8 @@ public final class GenericMultipleBarcodeReader implements MultipleBarcodeReader
       return;
     }
     boolean alreadyFound = false;
-    for (Result existingResult : results) {
-      if (existingResult.getText().equals(result.getText())) {
+    for (Result result : results) {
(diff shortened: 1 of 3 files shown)


srowen:
What's this tool?
I'd generally say this is too trivial to bother with, but it's really minor anyway.
Is there more?

core/src/main/java/com/google/zxing/multi/GenericMultipleBarcodeReader.java
```diff
@@ -86,8 +86,8 @@ private void doDecodeMultiple(BinaryBitmap image,
       return;
     }
     boolean alreadyFound = false;
-    for (Result existingResult : results) {
-      if (existingResult.getText().equals(result.getText())) {
+    for (Result result : results) {
```
srowen: Wait, this won't even compile. There's already a `result` in scope which is why this one is named differently.

condda: NATURALIZE is a framework that solves the coding convention inference problem for local conventions, offering suggestions to increase the stylistic consistency of a codebase.

Yes, there are more suggestions from Naturalize, you can go the website and download the tools available.

http://groups.inf.ed.ac.uk/naturalize/

Sorry about the compilation error, my mistake.

PR #412, opened 2015-07-02
Fixed substring call

strguntbr:
Fixed a bug when calling String.substring introduced in https://github.com/zxing/zxing/pull/395, which caused an exception when for example trying to encode "sadfg$sad" in PDF417 with the default encoding (ISO-8859-1)


srowen:
Do you have a moment to throw in a minimal unit test for this? if it looks like it would take more than 10 mins to adapt some of the existing code you can forget it and I can try to tack it on

strguntbr:
Done. The minimal test I added fails on master but passes on this branch.

Issue #418, opened 2015-07-15
PDF_417 lineThickness


awenger:
Is there a reason to start the PDF_417 generation with a lineThickness of 2 [1] if the barcode is later scalled [2]. This excludes the generation of barcodes with lineThickness of 3 for example.

[1] https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/pdf417/PDF417Writer.java#L100
[2] https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/pdf417/PDF417Writer.java#L121

srowen:
I dont' think I get the question (didn't write this code either) but if you can show and explain a change in a PR I can look at it.

awenger:
The current generation starts with lineThickness 2 and scales the barcode later to fit into the available space. This prevents for example the generation of a barcode with scale factor 1,3,5 etc.

PR #419, opened 2015-07-15
Set PDF_417 lineThickness to 1

awenger:
- fine-grained scaling for PDF_417 generation

#418


srowen:
That's probably OK. You likely know this code better than I. Is the effect just that it scales the result from 1-pixel bars rather than always starting at 2? that seems reasonable.

Issue #449, opened 2015-08-27
PDF417 decode error


gredler:
The following PDF417 bar code, which was generated using [Okapi](https://github.com/woo-j/OkapiBarcode/) and encodes the text *"This is just a test."* cannot be decoded by ZXing:

![normal-basic-6-rows-6-columns](https://cloud.githubusercontent.com/assets/178883/9533715/ef3d587e-4ce1-11e5-94bb-98def8edb995.png)

It decodes without any issues with a hardware bar code scanner.

I would attach a stack trace, but it throws a `FormatException` without a stack trace (at `DetectionResultRowIndicatorColumn#125`, in the `getRowHeights()` method).

graug:
The metadata cointained in the generated barcode does not match the actual barcode. According to the barcode metadata, the barcode is supposed to consist of 4 rows, but it actually consists of 6 rows. The last two rows of the barcode contain garbage. If I ignore the last two rows, the code decodes properly as "This is just a test.". I suggest you report this problem to Okapi.
If you cannot fix Okapi right away, you can apply the following patch to xing, which will cause it to ignore the additional garbage rows.

diff --git a/core/src/main/java/com/google/zxing/pdf417/decoder/DetectionResultRowIndicatorColumn.java b/core/src/main/java/com/google/zxing/pdf417/decoder/DetectionResultRowIndicatorColumn.java
index b9484c0..750aeac 100644
--- a/core/src/main/java/com/google/zxing/pdf417/decoder/DetectionResultRowIndicatorColumn.java
+++ b/core/src/main/java/com/google/zxing/pdf417/decoder/DetectionResultRowIndicatorColumn.java
@@ -122,7 +122,8 @@
       if (codeword != null) {
         int rowNumber = codeword.getRowNumber();
         if (rowNumber >= result.length) {
-          throw FormatException.getFormatInstance();
+        	continue;
+          //throw FormatException.getFormatInstance();
         }
         result[rowNumber]++;
       } // else throw exception?
diff --git a/core/src/main/java/com/google/zxing/pdf417/decoder/PDF417ScanningDecoder.java b/core/src/main/java/com/google/zxing/pdf417/decoder/PDF417ScanningDecoder.java
index 56ead37..50c78c1 100644
--- a/core/src/main/java/com/google/zxing/pdf417/decoder/PDF417ScanningDecoder.java
+++ b/core/src/main/java/com/google/zxing/pdf417/decoder/PDF417ScanningDecoder.java
[...]

PR #453, opened 2015-09-01
Ignore additional PDF417 barcode rows, https://github.com/zxing/zxing/issues/449

graug:
Some barcode generators (e.g. Okapi) seem to generate additional barcode
rows at the end of the barcode, which can be decoded as valid barcode
rows, but actually contain garbage. The barcode metadata states that
these additional rows are not part of the barcode. So we ignore those
addional rows when parsing the barcode. This could potentially also
happen (though quite unlikely) if data after the barcode looks like a
barcode row.


srowen:
Interesting, if it doesn't break detection of any existing barcodes, and doesn't introduce significantly false positives (highly unlikely) then this could be a good change. Right now it's failing the CI build because of a style problem -- tabs in one of the lines? https://travis-ci.org/zxing/zxing/jobs/78208877

graug:
Sorry about the formatting botch. I removed the tabs.

srowen:
OK, seems reasonable to me, thank you. I'm going to merge.

graug:
thx for the review and merging it.

PR #454, opened 2015-09-03
Support encoding UPC-E

RX1226:


core/src/main/java/com/google/zxing/oned/UPCEWriter.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.google.zxing.BarcodeFormat;
+import com.google.zxing.zxing.EncodeHintType;
+import com.google.zxing.zxing.WriterException;
+import com.google.zxing.zxing.common.BitMatrix;
+
+/**
+ * This object renders an UPC-E code as a {@link BitMatrix}.
+ *
+ * @author 0979097955s@gmail.com (RX)
```
srowen: You can add yourself to `AUTHORS` if you like.

srowen:
Check the Travis output -- looks like it doesn't build?

Is this basically a copy of the `EAN13Writer`? no problem if so, that's probably fine for such a short chunk of code.

RX1226:
This's basically a copy of the EAN13Writer, and I correct the import path.

srowen:
The only small issue I see is that you're using 4-space indent whereas the code uses 2-space.

Can you also squash these commits into one and force-push your branch? it would make the commit history tidier as otherwise they all go into the project history as is. It's possible for me to do it manually but takes some work.

RX1226:
Thanks you remind, I use 2-space instead of 4-space.

core/src/main/java/com/google/zxing/oned/UPCEWriter.java
RX1226: Cool!

srowen:
Are you able to squash commits too? Something roughly like

`git rebase -i HEAD~5`
Change all but the first "pick" to "squash" in an editor and save
Edit the commit messages down to one and save
`git push --force origin [your branch]`

I ask this when the commit history is not just 1-2 commits

RX1226:
Sorry I'm new to git, I don't understand how to do this operation. I afraid messed up this.

srowen:
If you have git on the command line, it's just about that simple as above -- try it out.
You can make a new PR from a patch generated from this one too.
I can also try to do it manually on this end if it's not working. It just takes some extra steps.
I can merge it as-is too but all 5 of your commits show up in the project history.

RX1226:
I create a new branch form zxing:master, but can't push it. I put the patch.diff to Dropbox, the path as following https://www.dropbox.com/s/ee5smc2074vzfwv/patch.diff?dl=0

srowen:
Yes you can't push to this repo; you push to your own repo's branch and open a pull request. You did that correctly already in this pull request.

I am happy to commit your patch myself if it's too much trouble; the only downside I suppose is it will appear that I authored the commit.

RX1226:
I success to branch and open a new pull requset. Thank you for your patience. :)

PR #461, opened 2015-09-14
Refactored openUrl()

andressantiago:
Added normalizeScheme() to replace manual normalization.

### DIFF
--- a/android/src/com/google/zxing/client/android/result/ResultHandler.java
+++ b/android/src/com/google/zxing/client/android/result/ResultHandler.java
@@ -421,14 +421,7 @@ public abstract class ResultHandler {
   }
 
   final void openURL(String url) {
-    // Strangely, some Android browsers don't seem to register to handle HTTP:// or HTTPS://.
-    // Lower-case these as it should always be OK to lower-case these schemes.
-    if (url.startsWith("HTTP://")) {
-      url = "http" + url.substring(4);
-    } else if (url.startsWith("HTTPS://")) {
-      url = "https" + url.substring(5);
-    }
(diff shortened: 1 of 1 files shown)


android/src/com/google/zxing/client/android/result/ResultHandler.java
```diff
@@ -421,14 +421,7 @@ final void searchBookContents(String isbnOrUrl) {
...
   final void openURL(String url) {
-    // Strangely, some Android browsers don't seem to register to handle HTTP:// or HTTPS://.
-    // Lower-case these as it should always be OK to lower-case these schemes.
-    if (url.startsWith("HTTP://")) {
-      url = "http" + url.substring(4);
-    } else if (url.startsWith("HTTPS://")) {
-      url = "https" + url.substring(5);
-    }
-    Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+    Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url).normalizeScheme());
```
srowen: Looks like this is 4.1+ only: http://developer.android.com/reference/android/net/Uri.html#normalizeScheme()  Right now it's still on 4.0.x. Eventually it'd be fine to up the minimum but I wouldn't do it just for this.

andressantiago:
Bizarre, somehow my compiler didn't pick that up. Completely agree, not worth the upgrade for this. Thanks for the speedy reply!

srowen:
I bet it's because the only Maven artifacts for Android 4 are for 4.1.1.4, so the Maven build compiles it fine. It might fail once you get to building the .apk. Thank you in any event.

PR #462, opened 2015-09-16
Add TIFF format with the help of ImageJ

loikki:
It seems that the java ImageIO is not able to read the TIFF format, I add the ImageReader from ImageJ (I put the dependency in maven. The new environment variable IMAGEJ needs to be create. The path is the following: /home/openerp//ImageJ/ij.jar). I think that maybe the ImageJ ImageReader should be used instead of the java one

### DIFF
--- a/javase/pom.xml
+++ b/javase/pom.xml
@@ -31,6 +31,11 @@
       <artifactId>jcommander</artifactId>
       <version>1.48</version>
     </dependency>
+    <dependency>
+      <groupId>net.imagej</groupId>
+      <artifactId>ij</artifactId>
+      <version>1.48r</version>
+    </dependency>
   </dependencies>
 
   <parent>
--- a/javase/src/main/java/com/google/zxing/client/j2se/ImageReader.java
+++ b/javase/src/main/java/com/google/zxing/client/j2se/ImageReader.java
@@ -23,6 +23,9 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URLDecoder;
+import ij.ImagePlus;
+import ij.io.Opener;
+
 
 /**
  * Encapsulates reading URIs as images.
@@ -47,7 +50,13 @@ public final class ImageReader {
       throw new IOException("Resource not found: " + uri, iae);
     }
     if (result == null) {
-      throw new IOException("Could not load " + uri);
+      Opener opener_ = new Opener();
+      ImagePlus tmp = opener_.openTiff(uri.toString().split(":")[1],1);
+      if (tmp == null) {
+        throw new IOException("Could not load " + uri.toString());
+      } else {
+        result = tmp.getBufferedImage();
+      }
     }
     return result;
   }
--- a/pom.xml
+++ b/pom.xml
@@ -78,6 +78,13 @@
         <version>4.12</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+	<groupId>imagej</groupId>
+        <artifactId>ij</artifactId>
+	<version>1.8.0_45</version>
+	<scope>system</scope>
+	<systemPath>${IMAGEJ}/ImageJ/ij.jar</systemPath>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 
--- a/pom.xml
+++ b/pom.xml
@@ -79,11 +79,9 @@
         <scope>test</scope>
       </dependency>
       <dependency>
-	<groupId>imagej</groupId>
-        <artifactId>ij</artifactId>
-	<version>1.8.0_45</version>
-	<scope>system</scope>
-	<systemPath>${IMAGEJ}/ImageJ/ij.jar</systemPath>
+	<groupId>net.imagej</groupId>
+	<artifactId>ij</artifactId>
+	<version>1.48r</version>
       </dependency>
     </dependencies>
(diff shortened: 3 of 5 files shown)


pom.xml
```diff
@@ -78,6 +78,13 @@
         <version>4.12</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+	<groupId>imagej</groupId>
+        <artifactId>ij</artifactId>
+	<version>1.8.0_45</version>
+	<scope>system</scope>
+	<systemPath>${IMAGEJ}/ImageJ/ij.jar</systemPath>
```
srowen: This won't work; it depends on a local artifact you downloaded. Is this not in Maven Central?

javase/src/main/java/com/google/zxing/client/j2se/ImageReader.java
```diff
@@ -47,7 +50,13 @@ public static BufferedImage readImage(URI uri) throws IOException {
       throw new IOException("Resource not found: " + uri, iae);
     }
     if (result == null) {
-      throw new IOException("Could not load " + uri);
+      Opener opener_ = new Opener();
```
srowen: I don't mind efforts to expand TIFF support. How about using JAI though? I'm a little less inclined to add a whole new dependency for it, but if we must, a "standard" Java lib might be preferably especially if it integrations with `ImageIO`.

srowen: Hm, JAI is really old, maybe not. I'm OK with a dependency if it's Maven, licensed reasonably, not too bloated. Also can you describe what kinds of TIFFs ImageIO doesn't read?

loikki: For the TIFF format, I have read a few forum and they say that you need another plugin (http://stackoverflow.com/questions/1954685/cant-read-and-write-a-tiff-image-file-using-java-imageio-standard-library). I am trying to read some 1-page scans (I am not allowed to show them). The code is only trying to read the first page (maybe as a future modification, the choice can be given to the user). I have looked at JAI but as you said, it is really old and by using ImageJ, a lot of image processing tools are imported (therefore they can be used in order to improve the detection).
For maven, it is my first time using it, I did not realize that you have a repository. I found ImageJ on it, I will add this dependency (and it will solve the other comment). Thank you for your comments.
For the license, the only thing that I found is on wikipedia: http://imagej.net/Licensing (it seems that the license is unclear due to the participation of non us government employee to a government code but it is in the public domain).
I do not know if the code is bloated (I did not looked at the source code) but the zip that you can download on their website is about 60MB.

javase/pom.xml
```diff
@@ -31,6 +31,11 @@
       <artifactId>jcommander</artifactId>
       <version>1.48</version>
     </dependency>
+    <dependency>
+      <groupId>net.imagej</groupId>
+      <artifactId>ij</artifactId>
+      <version>1.48r</version>
```
srowen: Looks like the newest version is 1.50b BTW?

srowen:
License is OK indeed. It's a 2MB dependency which is big but not crazy.
I'm not sure why but I'm a little unsure about adding this to all downstream consumers just to support a few more TIFFs, but it is useful.

Taking it to the extreme, I suppose this could be an optional dependency that's loaded only if present. Might be a bit overkill. At least maybe we could refactor the code a bit to try a series of approaches to load the image. Right now that's just two possibilities. But might be tidier and allow for more extension. Are you up for that, or, what is your opinion on the effort?

srowen:
Oh finally, probably worth adding a test image that only decodes when the dependency is added, just to make sure this code path is actually tested!

loikki:
As said before, I do not know maven well enough for this, but I think that we can try to load it only when an optional argument is given (if I have enough time, I will try to do it and the test image).
In any case, I still need to work on it for my job (I will try to give an optional argument in order to specify in which area is the barcode [I am thinking about giving a square], read PDF [I think that I will need to convert it before using the library but I can add an option in it] and add the possibility to read on a different page than the first one [really easy but I wanted to do some test before starting something like this]).

javase/pom.xml
loikki: I was using a older version, therefore I have put the first one, but yeah it is 1.50b

srowen:
That much I know, it can be a `<scope>optional</scope>` dependency. The code would have to load it by reflection ... meh, I'm still not sure whether I'm just over-thinking it. At least would be cool to refactor this into some kind of interface with two impls, if you're interested. Otherwise maybe I'll merge this and take a shot at it.

Could you squash and rebase your commits after updating the version?

loikki:
For the two impls, I won't have enough time and my boss won't allow it. For the squash and rebase, I will try (I still need to learn a few things about github).
I forgot to close the pull request before my last two commits, you do not need to look at them, it is still under writing (I am trying to give the option in CommandLineEncoder of the Error correction level [it is working for qr code, but it is awful]).

srowen:
OK NP. You can open a new PR. If you're short on time I can just commit your changes separately (and give you credit) and try some additional refactoring.

loikki:
I have enough time for doing something proper on github. As long as it is useful for my job, I can work on this code, but I won't be able to write a second implementation (at least as I am understanding it). I have done a few stupid thing with my fork, but I will create a proper pull request before tomorrow evening (European time).

PR #466, opened 2015-09-22
Add TIFF support

loikki:
Here is the good commit. Sorry about it

### DIFF
--- a/javase/pom.xml
+++ b/javase/pom.xml
@@ -31,6 +31,11 @@
       <artifactId>jcommander</artifactId>
       <version>1.48</version>
     </dependency>
+    <dependency>
+      <groupId>net.imagej</groupId>
+      <artifactId>ij</artifactId>
+      <version>1.48r</version>
+    </dependency>
   </dependencies>
 
   <parent>
--- a/javase/src/main/java/com/google/zxing/client/j2se/ImageReader.java
+++ b/javase/src/main/java/com/google/zxing/client/j2se/ImageReader.java
@@ -23,6 +23,8 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URLDecoder;
+import ij.ImagePlus;
+import ij.io.Opener;
 
 /**
  * Encapsulates reading URIs as images.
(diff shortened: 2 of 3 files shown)


javase/pom.xml
```diff
@@ -31,6 +31,11 @@
       <artifactId>jcommander</artifactId>
       <version>1.48</version>
     </dependency>
+    <dependency>
+      <groupId>net.imagej</groupId>
+      <artifactId>ij</artifactId>
+      <version>1.48r</version>
```
srowen: OK this is still the older version but I can fix that later. Is there a test image that isn't read by ImageIO? I probably need a test that shows this works.

loikki:
Sorry about this but now the code seems to work with TIFF (but if the barcode is not on the first page, it won't be read). I really do not understand why sometimes it works and sometimes it doesn't. I believe that java ImageIO has done an update in order to read the TIFF not so far ago and that sometimes my old version is called. It is really strange with all theses forum saying that the format is not readable.
If I find the reason, I will send you the information.

Sorry for loosing your time,
Best

srowen:
No problem. ImageIO depends on native libraries and so later versions of Java or the OS may improve support. If it possible the image was parsed but not decoded? Also if you are able to share an image offline I can try it, but maybe they have sensitive info.

loikki:
This is the TIFF that I wanted to use for the test in your library (I have changed the extension due to github that accepts only jpg,png and gif). I do not remember exactly the error, but it was something like "Exception in thread 'main' java.io.IOException: Could not load file:$FILE_NAME "
 
test - copie tiff

srowen:
I can't load this file with ImageIO on OS X + Java 8, so I think you do have a point. However, I found a nice solution. By just adding this plugin to ImageIO, it automatically works for me: https://github.com/jai-imageio/jai-imageio-core  I'll add this to `javase` and the you can try it again?

loikki:
Really strange, I have compiled again the source and now the code does not read TIFF format.
Nice way to do it, I will try your code as soon as possible.

loikki:
Seems to be perfect, Thank you very much :)

loikki:
Hi, I have a small question about your code before trying to do an other pull request.
I am looking for adding the error correction level (at least for the QR code) and I just want to be sure that only QR code have them.
Thank you in advance for your answer.
Best,

PR #469, opened 2015-09-28
Add choice of the correction error level for qr code

loikki:
Hi,
I have done the modification that I have spoken about, It add an option in the command line encoder for choosing the correction error level. I have chosen an integer to define it due to the different bar code format and their correction level class (the change to a char can be quickly done if needed).

### DIFF
--- a/javase/src/main/java/com/google/zxing/client/j2se/CommandLineEncoder.java
+++ b/javase/src/main/java/com/google/zxing/client/j2se/CommandLineEncoder.java
@@ -19,9 +19,13 @@ package com.google.zxing.client.j2se;
 import com.beust.jcommander.JCommander;
 import com.google.zxing.MultiFormatWriter;
 import com.google.zxing.common.BitMatrix;
+import com.google.zxing.EncodeHintType;
+
+import com.google.zxing.qrcode.decoder.ErrorCorrectionLevel;
 
 import java.nio.file.Paths;
 import java.util.Locale;
+import java.util.HashMap;
 
 /**
  * Command line utility for encoding barcodes.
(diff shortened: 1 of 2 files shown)


javase/src/main/java/com/google/zxing/client/j2se/CommandLineEncoder.java
```diff
@@ -47,8 +51,35 @@ public static void main(String[] args) throws Exception {
       outFileString += '.' + config.imageFormat.toLowerCase(Locale.ENGLISH);
     }
     
-    BitMatrix matrix = new MultiFormatWriter().encode(
+    BitMatrix matrix = null;
```
srowen: Yikes, this is a lot of extra code to special case one arg to one encoder. Why not handle this via the existing hint? It wouldn't hurt to let the code downstream parse an String if needed to make this work. No need for an arbitrary mapping to int

loikki:
I have the same feeling about this code, but I wasn't able to convert the integer to the good class with the hint (I don't have enough experience in java to do it properly). I tried with a different approach more compact but java (with my knowledge of this language) is not permissive enough to do it. The first version was in QRCodeWriter, I was checking the class and if it was an int, I was trying to transform it to a ErrorCorrectionLevel but I don't know how to do it. If you have any suggestion, I can rewrite it, otherwise this version is enough for my project.

srowen:
Actually, I think that if you tell JCommander that the type is an enum type like ErrorCorrectionLevel, it will know to parse the string value of the flag as an enum for you. Then it's just an object you have to put into a hint Map under key EncodeHintType.ERROR_CORRECTION. Pass that to MultiFormatWriter.encode() as well and I think it works.

loikki:
The problem that I have seen with this approach is with the other kind of bar code that do not have the same kind of ErrorCorrectionLevel (this function is a specific thing for QR code and does not have the same kind of class). I will do a test with your method in a few days

srowen:
You raise a good point. Maybe it should after all be a String, and then, downstream code needs to change to handle the possibility of a String-valued hint for this (and other) types. How about that? then the generic code in the encoder stays quite generic.

srowen:
@loikki I added support for String arguments to most encode hints, including this one.
https://github.com/zxing/zxing/commit/a6707cb4407ebe771e1b81ae2cab0e136c548b25

loikki:
Thank you very much
Sorry for not taking care of it, I had a lot of thing to do

PR #472, opened 2015-10-01
fix a little error

qianchenglenger:
RGB to gray, the weight of R is 0.299, not 0.229.  The comment have a little error. And why add 0x200 ? It may be a error.  306/1024 = 0.2988.  601/1024=0.5869 and 117/1027 = 0.1143.


srowen:
You're right about the comment.
The constant factor there implements rounding, effectively, as in https://en.wikipedia.org/wiki/YUV#Full_swing_for_BT.601
You can add this to the comment as well.

There are two copies of BufferedImageLuminanceSource actually, unfortunately, for dependency reasons. Update both?

PR #476, opened 2015-10-12
Add possibility to choose the error correction level for the encoding

loikki:


loikki:
Hi,
As I told you last week, I have done the option to choose the correction level (using a string)

javase/src/main/java/com/google/zxing/client/j2se/EncoderConfig.java
```diff
@@ -48,6 +48,10 @@
       validateWith = PositiveInteger.class)
   int height = 300;
 
+  @Parameter(names = "--error_correction_level",
+      description = "Error correction level for the encoding")
+  String errorCorrectionLevel = "";
```
srowen: Tiny, but let's just use `null` as a default

javase/src/main/java/com/google/zxing/client/j2se/CommandLineEncoder.java
```diff
@@ -46,9 +48,13 @@ public static void main(String[] args) throws Exception {
     if (EncoderConfig.DEFAULT_OUTPUT_FILE_BASE.equals(outFileString)) {
       outFileString += '.' + config.imageFormat.toLowerCase(Locale.ENGLISH);
     }
-    
+ 
+    HashMap<EncodeHintType,String> hints = new HashMap<EncodeHintType,String>();
```
srowen: Nits: use `Map` as the variable type, and omit the generic types after `HashMap`. You're also missing spaces before arguments as in ",hints)" instead of ", hints)"

loikki: I have done the changes (except the part with the generic type, because I do not understand it and I do not know which HashMap you are speaking of). What should I change in the following line:
Map<EncodeHintType,String> hints = new HashMap<EncodeHintType,String>();

Thank you in advance

srowen: Oh I just mean write `= new HashMap<>();`

loikki:
I have done the changes and merge everything in one commit

javase/src/main/java/com/google/zxing/client/j2se/CommandLineEncoder.java
```diff
@@ -19,9 +19,12 @@
 import com.beust.jcommander.JCommander;
 import com.google.zxing.MultiFormatWriter;
 import com.google.zxing.common.BitMatrix;
+import com.google.zxing.EncodeHintType;
 
 import java.nio.file.Paths;
 import java.util.Locale;
+import java.util.Map;
+import java.util.HashMap;
```
srowen: I'm sorry to nit pick, but while we're here, there are a number of style issues. (Some IDEs can help with this.)

Imports are not ordered alphabetically; still missing a space after arg on line 55; line 58 adds an extra indent that should not be there, and the type of `hints` can really be `Map<EncodeHintType,Object>` as it does not in general only map to strings.

loikki: ok, It is not a problem, I can understand that.
I am downloading eclipse and correcting everything

loikki:
I have done the few improvements

PR #486, opened 2015-10-21
Create a specific DataMatrixWriter to generate 2D-Doc

gbm-bailleul:
2D-Doc is a specific usage of Datamatrix defined in France, see [1] for more information.

There is no issue to read 2D-Doc with zxing.

The only issue is about generation because 2D-Doc only uses ASCII encoding and does not optimize the encoding.

This patch:
* modify HighLevelEncoder to allow bypass of encodation optimization
* provide a new specific writer: DDDocWriter
* provide a 2D-Doc encoder
* provide some local changes to glue all the stuff

This is my first attempt with zxing, so I am waiting for your feedbacks to improve this proposal.

KR,
Guillaume

[1] https://ants.gouv.fr/content/download/516/5665/version/4/file/ANTS_2D-Doc_CABSpec_v2.0.1_erratum.pdf
in french, sorry

### DIFF
--- a/core/src/main/java/com/google/zxing/datamatrix/DDDocWriter.java
+++ b/core/src/main/java/com/google/zxing/datamatrix/DDDocWriter.java
@@ -1,0 +1,10 @@
+package com.google.zxing.datamatrix;
+
+import com.google.zxing.datamatrix.encoder.HighLevelEncoder;
+
+public class DDDocWriter extends DataMatrixWriter {
+
+    public DDDocWriter(){
+        super(HighLevelEncoder.DDDOC_ENCODATION);
+    }
+}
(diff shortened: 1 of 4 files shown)


srowen:
Hm, my concern here is just about adding code to support a very niche format. It's not much, but also I am not clear how much this is used if ever.

gbm-bailleul:
I agree that this format looks niche right now. The interest in France is growing but not so quickly. According to the french gov service in charge of this format, there is some interest on that in Germany, but I would not give any value to that information.

My first attempt was to only extend the existing DatamatrixWriter but the way HighLevelEncoder is done blocked  me.

Maybe a better idea would be to propose some rework on HighLevelEncoder to make it more configurable and make specific encoder injection possible.

What do you think ?

PR #508, opened 2015-12-02
QRCode Encoder: Support NUMERIC/ALPHANUMERIC Mode in Shift_JIS

yoshi389111:
I've used a QRCode Encoder.
When Charset is Shift_JIS, byte mode has been chosen in the content of numeric data.

```
    Map<EncodeHintType,Object> hints = new EnumMap<>(EncodeHintType.class);
    hints.put(EncodeHintType.CHARACTER_SET, "Shift_JIS");
    QRCode qrCode = Encoder.encode("0123", ErrorCorrectionLevel.M, hints);
    System.out.println(qrCode.getMode()); // -> BYTE
```

I think should be chosen numeric mode.


srowen:
Yes I think that's probably fine. Although someone who specifically selects Shift_JIS is probably intending to encode Japanese characters, and this will generally require byte mode, I agree that this doesn't mean numeric mode is not possible if the input is numbers.

yoshi389111:
Thank you very much. :smile:

PR #542, opened 2016-01-20
Fixed coding bug in QRcode encoding mask penalty rule #3 calculation

erappos:
A bug exists in the way zxing calculates the mask to apply during QR code generation.

Eight masks have to be tested and the one with the lowest penalty is chosen to be applied to the QR code being created. Each mask is scored on four criteria (rules), and rule 3 gives a penalty based on a particular sequence of black and white squares being present in the code. 

zxing uses two helper functions ```isWhiteHorizontal``` and ```isWhiteVertical```  to test if a given range of cells are all white (=0). 

However, a bug in the functions makes them return ```true``` even if the range of cells is out of range. This results often in higher rule 3 penalty values and can end up with a QR code choosing a suboptimal masking pattern.

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/encoder/MaskUtil.java
+++ b/core/src/main/java/com/google/zxing/qrcode/encoder/MaskUtil.java
@@ -107,6 +107,9 @@ final class MaskUtil {
       if (i >= 0 && i < rowArray.length && rowArray[i] == 1) {
         return false;
       }
+      if(i<0 || i >= rowArray.length){
+        return false;
+      }
     }
     return true;
   }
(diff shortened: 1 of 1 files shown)


srowen:
I haven't read the code in a while, but I believe the assumption is that anything outside the QR code's boundary is white. This seems to implement that correctly. Is this wrong according to the spec? could be, I haven't looked at this in ages.

This could be written more simply I think if so, by checking the range upfront rather than in the loop.

erappos:
Perhaps... not clear in the official specification what is considered "existance of the pattern in the symbol".  So best to leave as is then due to lack of certainty.   Thanks

srowen:
Yeah, I'm not entirely sure either, but if there's evidence that the interpretation should change, that's fine. As you can see it makes some tests fail, but the tests could be wrong.

PR #549, opened 2016-01-29
Added the "Get Started Developing" link in the README.md

az-ericso:
It is really hard to find the instruction.


srowen:
Yes the new layout no longer links wikis on the front page. I am fine with this but I think it should reference getting started _developing_

az-ericso:
I already did.  Is that the wrong link?

srowen:
The text you added just talks about 'getting started' but this has only to do with development

az-ericso:
how about now?

PR #389, opened 2015-05-14
Added rawBytes functionality to the Aztec scanner

iSDP:
Added rawBytes functionality to the Aztec scanner

Issue #381, opened 2015-05-13
Aztec Decoder broken, getRawBytes never populated


iSDP:
I've found the issue why getRawBytes() is not populated, it can be found [here](https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java#L79)

What needs to be done to pass the raw Bytes along?

iSDP:
Fixed it myself, do you want the code?

srowen:
Yeah, like turning the 'bits' into a `byte[]`? I think that was never implemented because the code came from a different place. The bytes may or may not be meaningful, but better to return something than nothing. If you have a PR for that, that's fine.

iSDP:
I'll look into it

iSDP:
Made a PR: https://github.com/zxing/zxing/pull/389

alotaibiatm:
Tanks

PR #583, opened 2016-04-22
cleaned up the new aztec decoder by @iSDP referenced in issue #381

oelna:
A while back @iSDP complained that the Aztec decoder does not return raw bytes like eg. the QR decoder. He made a PR [#389](https://github.com/zxing/zxing/pull/389), but it was rejected due to formatting issues.

I'd really like for this to work. Many Aztec barcodes don't store string data but compressed zlib streams, which get corrupted when turned into strings. The raw bytes would help.

I have not written any of the lines in this PR, only reformatted @iSDP's work to match the repo's conventions. I hope this is ok.

### DIFF
--- a/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
+++ b/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
@@ -338,6 +338,8 @@ public final class Decoder {
     for (int i = 0; i < 8; i++) {
       if (array[start + i]) {
         b |= 1 << (7 - i);
+      } else {
+        throw new IndexOutOfBoundsException();
       }
     }
     return b;
--- a/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
+++ b/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
@@ -336,16 +336,18 @@ public final class Decoder {
   private static byte boolsToByte(final boolean[] array, final int start) {
     byte b = 0;
     for (int i = 0; i < 8; i++) {
-      if (array[start + i])
+      if (array[start + i]) {
         b |= 1 << (7 - i);
+      }
     }
     return b;
   }
 
   public static byte[] convertBoolArrayToByteArray(boolean[] boolArr) {
     byte[] byteArr = new byte[boolArr.length / 8];
-    for (int i = 0; i < byteArr.length; i++)
+    for (int i = 0; i < byteArr.length; i++) {
       byteArr[i] = boolsToByte(boolArr, 8 * i);
+    }
     return byteArr;
   }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
```diff
@@ -332,6 +333,22 @@ private static int readCode(boolean[] rawbits, int startIndex, int length) {
...
+    byte b = 0;
+    for (int i = 0; i < 8; i++) {
+      if (array[start + i])
+        b |= 1 << (7 - i);
+    }
+    return b;
+  }
+
+  public static byte[] convertBoolArrayToByteArray(boolean[] boolArr) {
+    byte[] byteArr = new byte[boolArr.length / 8];
```
srowen: Is it certain that the array's length will be a multiple of 8? otherwise this would not work.
Have you tested the result to see that it matches what you expect? Maybe piggy-backing a check for this result on some of the existing tests would be a good idea.
Finally, add braces for all if and for statcements.

iSDP:
It works as expected: https://m.facebook.com/brainblendr/photos/a.748274335293783.1073741829.726550520799498/748274018627148/?type=3&source=54

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
oelna: I certainly can add the missing braces and resubmit, but I don't feel comfortable changing the actual code, as it's a bit over my head. Maybe @iSDP can weigh in on this or add the checks you're talking about.

srowen: At least throw an exception if this condition isn't true, otherwise some data is silently ignored. Add at least one test and this is good to merge.

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
```diff
@@ -332,6 +333,26 @@ private static int readCode(boolean[] rawbits, int startIndex, int length) {
     return res;
   }
 
+  private static byte boolsToByte(final boolean[] array, final int start) {
+    byte b = 0;
+    for (int i = 0; i < 8; i++) {
+      if (array[start + i]) {
+        b |= 1 << (7 - i);
+      } else {
+        throw new IndexOutOfBoundsException();
```
srowen: What's this about? it certainly isn't an error

oelna: I feel @iSDP and I have hit a dead end here, as we both don't know enough Java to fix this or throw a decent error. Until somebody figures this out, Or I learn enough, there is nothing I can do. Sorry about the trouble and thanks for listening.

srowen: Ok. Just mean I am not sure why you added this change. It is wrong and makes the test fail. I suggested checking the length of the array. Containing false is not the same thing nor an error. It also needs a simple test to demonstrate and verify the expected behavior

tanelihuuskonen:
Just in case you didn't notice, I wrote a fix for this.  I'm afraid I forgot to mention your PR in mine. #591

oelna:
I'm so happy we can put this to rest. Thank you for helping out!
(I'm actually hoping this change makes it into ZXingObjC eventually) 😊

PR #584, opened 2016-04-23
Text oriented code

movedon2otherthings:
This pull request is a strawman proposal about a novel barcode type TO code.
The actual implementation currently is an one java class decoder, baked into the demo android app.
The decoder:
https://gist.github.com/movedon2otherthings/7d3052473e1c228e5ad389845aa7ccb0

I need an official BarcodeFormat assigned, because as you can see at line 728 I currently use BarcodeFormat.UPC_E, which is not nice.

Yes, this is original work.
Yes, this is open source.

### DIFF
--- a/core/src/main/java/com/google/zxing/BarcodeFormat.java
+++ b/core/src/main/java/com/google/zxing/BarcodeFormat.java
@@ -64,6 +64,9 @@ public enum BarcodeFormat {
 
   /** RSS EXPANDED */
   RSS_EXPANDED,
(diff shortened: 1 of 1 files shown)


srowen:
No, I definitely would not add this to the main project if it's a non-existent type. It wouldn't help because there's no code to decode whatever this is. However you can modify your fork however you like.

movedon2otherthings:
Thank you for letting me know.

PR #589, opened 2016-05-04
Polish corrections

tanelihuuskonen:
I attempted to correct some Polish translations.  I'm fairly sure my first commit actually fixes two errors, but I don't recommend accepting the second one without a native speaker's approval (and probably some corrections to my corrections).

I hereby grant everyone the right to use and distribute this contribution under the same terms as the original program.


srowen:
Thanks!

PR #591, opened 2016-05-05
Aztec decoder returns raw bytes

tanelihuuskonen:
The Aztec decoder now packs the code bits into bytes and returns them as "raw bytes", even though the error correction codes and stuffing have been already removed.  Also added a couple of test cases for the byte-packing function.  This fixes issue #381.

I hereby grant everyone the right to use and distribute this contribution under the same terms as the original program.

### DIFF
--- a/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
+++ b/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
@@ -347,7 +347,7 @@ public final class Decoder {
   /**
    * Packs a bit array into bytes, most significant bit first
    */
-  public static byte[] convertBoolArrayToByteArray(boolean[] boolArr) {
+  static byte[] convertBoolArrayToByteArray(boolean[] boolArr) {
     byte[] byteArr = new byte[(boolArr.length + 7) / 8];
     for (int i = 0; i < byteArr.length; i++) {
       byteArr[i] = readByte(boolArr, 8 * i);
--- a/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
+++ b/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
@@ -75,8 +75,9 @@ public final class Decoder {
     BitMatrix matrix = detectorResult.getBits();
     boolean[] rawbits = extractBits(matrix);
     boolean[] correctedBits = correctBits(rawbits);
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
```diff
@@ -332,6 +333,28 @@ private static int readCode(boolean[] rawbits, int startIndex, int length) {
...
+    if (n >= 8) {
+      return (byte) readCode(rawbits, startIndex, 8);
+    }
+    return (byte) (readCode(rawbits, startIndex, n) << (8 - n));
+  }
+
+  /**
+   * Packs a bit array into bytes, most significant bit first
+   */
+  public static byte[] convertBoolArrayToByteArray(boolean[] boolArr) {
```
srowen: Can be package-private?

core/src/test/java/com/google/zxing/aztec/decoder/DecoderTest.java
```diff
@@ -105,4 +106,35 @@ public void testDecodeTooManyErrors2() throws FormatException {
...
+  private static void assertEqualByteArrays(byte[] b1, byte[] b2) {
+    assertEquals(b1.length, b2.length);
+    for (int i = 0; i < b1.length; i++) {
+      assertEquals(b1[i], b2[i]);
+    }
+  }
+
+  @Test
+  public void testRawBytes() {
+    boolean bool0[] = {};
```
srowen: Nit: `boolean[] bool0...` (Java-style array declarations)

core/src/test/java/com/google/zxing/aztec/decoder/DecoderTest.java
```diff
@@ -105,4 +106,35 @@ public void testDecodeTooManyErrors2() throws FormatException {
     new Decoder().decode(r);
   }
 
+  private static void assertEqualByteArrays(byte[] b1, byte[] b2) {
```
srowen: JUnit already has ... http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertArrayEquals(byte[], byte[])

tanelihuuskonen:
Thanks for the comments.  I only started learning Java a couple of months ago, so my C habits keep showing up.  I just committed my corrections.

PR #596, opened 2016-05-11
Simplified some code

tanelihuuskonen:
Simplified a few bits of code in BitArray and RGBLuminanceSource, improving performance at least marginally.

### DIFF
--- a/core/src/main/java/com/google/zxing/common/BitArray.java
+++ b/core/src/main/java/com/google/zxing/common/BitArray.java
@@ -151,7 +151,7 @@ public final class BitArray implements Cloneable {
    * @param end end of range, exclusive
    */
   public void setRange(int start, int end) {
-    if (end < start) {
+    if (end < start || start < 0 || end > size) {
       throw new IllegalArgumentException();
     }
     if (end == start) {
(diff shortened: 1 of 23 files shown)


core/src/main/java/com/google/zxing/common/BitArray.java
```diff
@@ -163,15 +163,8 @@ public void setRange(int start, int end) {
...
-      if (firstBit == 0 && lastBit == 31) {
-        mask = -1;
-      } else {
-        mask = 0;
-        for (int j = firstBit; j <= lastBit; j++) {
-          mask |= 1 << j;
-        }
-      }
+      // Ones from firstBit to lastBit, inclusive
+      int mask = (2 << lastBit) - (1 << firstBit);
```
srowen: That's pretty cool, never though of writing it this way. Seems to work. You can inline this in the line below if you like.

srowen:
Yeah looks like nice optimizations, well done. I'll merge it when I get back.

core/src/main/java/com/google/zxing/common/BitArray.java
tanelihuuskonen: Right, that makes sense.  Done.

tanelihuuskonen:
Thanks.  I like that kind of stuff.

PR #601, opened 2016-05-24
Simplified pieces of code

nayuki:
I simplified a couple of pieces of code while keeping the behavior the same. The message in each commit explains what happened.

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/decoder/DataMask.java
+++ b/core/src/main/java/com/google/zxing/qrcode/decoder/DataMask.java
@@ -139,16 +139,4 @@ enum DataMask {
 
   abstract boolean isMasked(int i, int j);
 
-  /**
-   * @param reference a value between 0 and 7 indicating one of the eight possible
-   * data mask patterns a QR Code may use
-   * @return DataMask encapsulating the data mask pattern
-   */
-  static DataMask forReference(int reference) {
-    if (reference < 0 || reference > 7) {
-      throw new IllegalArgumentException();
-    }
-    return values()[reference];
-  }
-
 }
--- a/core/src/main/java/com/google/zxing/qrcode/decoder/DataMask.java
+++ b/core/src/main/java/com/google/zxing/qrcode/decoder/DataMask.java
@@ -29,24 +29,96 @@ import com.google.zxing.common.BitMatrix;
  *
  * @author Sean Owen
  */
-abstract class DataMask {
+enum DataMask {
 
   /**
    * See ISO 18004:2006 6.8.1
    */
-  private static final DataMask[] DATA_MASKS = {
-      new DataMask000(),
-      new DataMask001(),
-      new DataMask010(),
-      new DataMask011(),
-      new DataMask100(),
-      new DataMask101(),
-      new DataMask110(),
-      new DataMask111(),
+
+  /**
+   * 000: mask bits for which (x + y) mod 2 == 0
+   */
+  DATA_MASK_000() {
+    @Override
+    boolean isMasked(int i, int j) {
+      return ((i + j) & 0x01) == 0;
+    }
+  },
+
+  /**
+   * 001: mask bits for which x mod 2 == 0
+   */
+  DATA_MASK_001() {
+    @Override
+    boolean isMasked(int i, int j) {
+      return (i & 0x01) == 0;
+    }
+  },
+
+  /**
+   * 010: mask bits for which y mod 3 == 0
+   */
+  DATA_MASK_010() {
+    @Override
+    boolean isMasked(int i, int j) {
+      return j % 3 == 0;
+    }
+  },
+
+  /**
+   * 011: mask bits for which (x + y) mod 3 == 0
+   */
+  DATA_MASK_011() {
+    @Override
+    boolean isMasked(int i, int j) {
+      return (i + j) % 3 == 0;
+    }
+  },
+
+  /**
+   * 100: mask bits for which (x/2 + y/3) mod 2 == 0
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/google/zxing/qrcode/encoder/MatrixUtil.java
```diff
@@ -271,12 +271,7 @@ static void embedDataBits(BitArray dataBits, int maskPattern, ByteMatrix matrix)
   // - findMSBSet(1) => 1
   // - findMSBSet(255) => 8
   static int findMSBSet(int value) {
-    int numDigits = 0;
-    while (value != 0) {
-      value >>>= 1;
-      ++numDigits;
-    }
-    return numDigits;
+    return 32 - Integer.numberOfLeadingZeros(value);
```
srowen: Use highestOneBit? The only concerns are not using anything that is not present in Java 7 and Android's 4.0.x API but the two you use appear fine.

core/src/main/java/com/google/zxing/qrcode/encoder/MaskUtil.java
```diff
@@ -76,24 +76,24 @@ static int applyMaskPenaltyRule3(ByteMatrix matrix) {
...
         }
         if (y + 6 < height &&
-            array[y][x] == 1  &&
-            array[y +  1][x] == 0  &&
-            array[y +  2][x] == 1  &&
-            array[y +  3][x] == 1  &&
-            array[y +  4][x] == 1  &&
-            array[y +  5][x] == 0  &&
-            array[y +  6][x] == 1 &&
+            array[y + 0][x] == 1 &&
```
srowen: Space changes are OK. Maybe trivial but I don't like the + 0 no op just to make things line up

core/src/main/java/com/google/zxing/qrcode/decoder/DataMask.java
```diff
@@ -29,135 +29,126 @@
...
+  }
+
+  abstract boolean isMasked(int i, int j);
+
+  /**
+   * @param reference a value between 0 and 7 indicating one of the eight possible
+   * data mask patterns a QR Code may use
+   * @return DataMask encapsulating the data mask pattern
+   */
+  static DataMask forReference(int reference) {
```
srowen: I was going to say we can remove this method because enum exposes values(). But it does very technically change the exception for bad values. Still it is an internal API and really this error condition ought to be caught by the framework. I'm ok removing it and updating all callers to use values()

srowen:
Yeah nice simplifications. Lots of this code comes from the J2ME era so was much more limited in what it could call.

core/src/main/java/com/google/zxing/qrcode/encoder/MatrixUtil.java
nayuki: I know for a fact that the two functions are in Java 5. I checked the Android API now as per your suggestion, and it looks fine too: https://developer.android.com/reference/java/lang/Integer.html#bitCount%28int%29 (API level 1, Android platform 1.0). =)

core/src/main/java/com/google/zxing/qrcode/encoder/MaskUtil.java
nayuki: Yeah, these changes are pretty taste-dependent. You might find the spacing changes to be too trivial to accept, and object to the `+ 0` no-op. I happen to do both in my coding projects, but this is your project and I'll leave the discretion to you.

core/src/main/java/com/google/zxing/qrcode/decoder/DataMask.java
nayuki: Okay, understood. Will you or I make the additional changes?

nayuki:
J2ME - that makes for an interesting project history. Did some incarnation of the ZXing barcode scanner app run on pre-Android phones? Or was ZXing acting as a library for an app to build on top of?

core/src/main/java/com/google/zxing/qrcode/decoder/DataMask.java
srowen: Please go ahead if you are willing to.

nayuki:
I'm not sure which way you would like to proceed with the pull request. If you'd like, feel free to cherry-pick or edit my commits, or put new ones on top to make corrections. Or if you'd like me to change what I did and submit new commits, please let me know. Thanks!

srowen:
It predated Android and so the very earliest versions supported J2ME. Here is some history: http://qr.ae/8VkwBg

srowen:
I'd appreciate it if you make additional changes discussed here, if you have the time. Then I will merge.

nayuki:
Added new commits to actualize the suggestions.

core/src/main/java/com/google/zxing/qrcode/encoder/MatrixUtil.java
srowen: What about using highestOneBit here instead?

nayuki: Nope, that would return the 1 bit in that position instead of the number of bit places. For example:

    0b000010 -> 0b000010
    0b000101 -> 0b000100
    0b001111 -> 0b001000

(My apologies for missing that half of your comment the first time.)

srowen: Ah OK. Doing this remotely on my phone or else would have properly investigated. Looks good.

PR #603, opened 2016-05-25
 Added some refactoring to the BitMatrix.java class

dannyhorvath1992:
Added some refactoring to the BitMatrix.java class


srowen:
Hm, I think this is a lot of change and some may not be worth it. For example 32 is so clearly understood as bits in an int and can never change, so not sure it helps to make it a constant. The name would be all caps anyway and I don't think it is a maximum number of bits. I also understand expanding pos to position but I don't think it adds readability net net. There are some indentation issues and some accidentally changed javadoc I think. BitRegion is interesting but it requires making an object every time methods like this are called internally and I think that adds nontrivial overhead for not much readability gain. Maybe there is another way to do that without object creation, to at least factor out at checking. As it is I would prefer not to make these changes.

PR #616, opened 2016-06-23
Improve efficiency of DataMatrix detector

IliasStavrakis:
This pull request improves efficiency of DataMatrix detector in TRY_HARDER mode. Using existing image sets the DataMatrixBlackBox2TestCase decodes 6 more images in 0 degrees rotation, 1 more in 180 degrees rotation and 1 more in 270 degrees rotation.

The main motivation for the proposed changes was the poor decode results of DataMatrix reader having as input scanned images that contain DataMatrix barcodes. We had large sets of scanned images that have little distortion caused by the scanner or pencil lines by humans on the barcode. In this version we can read images like the following:

cropimage1

githubimage2

The main enhancements in detector algorithm are the following:
1.We don’t need to correct TopRight edge point, the edges are computed correctly from the beginning thus the initial computation of dimension does not need correction.
2.In order to compute transitions we sample more segments, moving edge points in proper direction. The final result is the most popular transition among samples.
3.To compute edge points in a rectangle we enhance the existing algorithm that uses getBlackPointOnSegment() with two more segments. The one moves on horizontal axis at double speed, and the other moves on vertical axis at double speed. Finally, we combine the results to compute the edge point.
4.We introduce tolerance in WhiteRectangleDetector. The result is that the algorithm is more efficient in low quality images with distortion. 
5.We enhance the calculation of transitions of a segment with a validation algorithm that compares each transition length to the length of the most popular transition. If a length is too small (probably sign of distortion of human interpolation) we investigate the problematic area and if find proofs of distortion we correct the calculation of transition.
Finally, there are no changes in algorithm in normal mode (that does not uses TRY_HARDER hint).

### DIFF
--- a/core/src/main/java/com/google/zxing/common/detector/WhiteRectangleDetector.java
+++ b/core/src/main/java/com/google/zxing/common/detector/WhiteRectangleDetector.java
@@ -34,6 +34,7 @@ public final class WhiteRectangleDetector {
 
   private static final int INIT_SIZE = 10;
   private static final int CORR = 1;
+  private static final int TOLERANCE = 2;
 
   private final BitMatrix image;
   private final int height;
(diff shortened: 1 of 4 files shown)


srowen:
Generally that's a good improvement. It is a lot of added code and much of it duplicated. I wonder if it can be refactored? There are also a fair number of variations from the standard Java style of the rest of the code, like spaces around braces and so on.

IliasStavrakis:
I refactored code to remove duplication. For style issue it would be helpful to add check style constraints to better understand at build time the code that does not mach with the rest. Otherwise, could you show me a sample line and I will refactor the rest.

srowen:
OK, I made some additional changes to the checkstyle config that will flag the whitespace changes among other things. I'll look at the rest.

srowen:
Hm, I'm conflicted about this. The performance improvement is good though it will not affect regular scanning use cases. It's a fair bit of extra code complexity even now though.

Jaykob:
No offence, but can you explain a bit more in detail why an obvious performance improvement wasn't merged? It seems that regular scanning use cases weren't affected and the whole thing would only be invoked with the TRY_HARDER flag. And isn't this exactly what TRY_HARDER is supposed to do?

Browsing through the various datamatrix conversations I can't help myself but thinking there are commercial interests behind this trying to prevent ZXing from becoming too good in this heavily commercialized area? 
Just my 2 cents ;-)

IliasStavrakis:
Do you want the truth or a "politically correct" answer? You have already answered the question. Image at that period I worked for a team that had several more ideas to improve ZXing but these ideas blocked and now are dead when we understood what happened with this obvious improvement.

Jaykob:
Sad story and thanks for confirming my impression... 👎
Would be interesting to know the details in this case. A big 💰 was on the table somewhere for sure.

Anyways, I'm on the search for a C++ lib that would at least give me something to start with. Is there something you could recommend me?

srowen:
@Jaykob your comments are uninformed and strange. There's nothing hidden here. Nobody works on this project now, except that I try to look at bug fixes and do a release once a year. Nobody has worked on it full time for 9 years. There has never been any commercial interest in or support of this project. All volunteers. And it's long been in maintenance-only mode.

@IliasStavrakis I am not sure what you're talking about -- not sure what either of you are talking about. As I say above this is a ton of extra code for marginal improvement. If it were my day job to work on this I still wouldn't probably think this is worth it. As nobody is on hand to commit to support your code, I declined it.

Please, do what you want in your own fork.

PR #631, opened 2016-07-07
Update PreviewCallback.java

zhaoshuyu:
has a synchronize problem in a high performence device, need set previewHandler to null first


tanelihuuskonen:
Did you notice that your pull request failed an automatic test?

srowen:
What problem would this solve though? it doesn't really change anything from a sync/memory model perspective. (Failure was just a style check failure, but would have to be fixed too.)

PR #637, opened 2016-07-23
Add ability to specify a QR code version hint.

germinator:
Add a hint to allow the ability to specify a specific version of QR code. If the specified content won't fit in the version, an exception is thrown. If no hint is provided, the current method of selecting the "smallest" version that fits is used.

(I wanted to be able to generate a replica of a code that I had the content for.)


srowen:
What is the use case for making the version higher than needed?

germinator:
I wanted to be able to generate an exact replica of a code that I had content for. As to why the original creator of the code made the choice of a higher version than needed; I couldn't say.

srowen:
OK, looks good. There are some minor formatting / style things I'd like to adjust but I can do that separately.

Issue #642, opened 2016-08-01
Core: Aztec decoding incorrect for codes with U/S B/S


fgrieu:
ZXing (as of July 25, 2016) decodes the following Aztec code as 3333h3i3j33333, when that should I believe be 3333h3i3jITIT per a literal reading of the ISO/IEC 24778:2008 standard, and is what I get with at least one popular decoder independent of ZXing (NeoReader).
![dlusbs.png](https://cloud.githubusercontent.com/assets/7809433/17298916/4fe36296-580c-11e6-8345-6ba524b7d232.png)

The binary sequence in that code, before applying the process of _14.5.2 Interpreting the bit stream_ in the standard, is the bit sequence 11110010101010101010111111111100101011010000011001101101001001100110110101001010101010101010101 which parses to
```
  per ZXing             per the standard and Neoreader
11110    (D/L)             11110    (D/L)
0101      3                0101      3
0101      3                0101      3
0101      3                0101      3
0101      3                0101      3
1111      (U/S)            1111      (U/S)
11111     (B/S)            11111     (B/S)
00101     [5]              00101     [5]
01101000  h                01101000  h
00110011  3                00110011  3
01101001  i                01101001  i
00110011  3                00110011  3
01101010  j                01101010  j
0101      3                01010     I
0101      3                10101     T
0101      3                01010     I
0101      3                10101     T
0101      3
```
The discrepancy boils down to diverging interpretation of :
> Byte Shift (B/S) is a special case, shifting into a run length-controlled string of literal 8-bit bytes. Following a B/S is a 5-bit binary value: if non-zero, it encodes the number of bytes (1 to 31) that follow, but if zero then the next 11 bits encode the number of bytes less 31. Byte Shift can thus encode either isolated extended ASCII or control characters or long strings of byte data, possibly filling the whole symbol. At the end of the byte string, **encoding returns to the _mode_ from which B/S was invoked**.

The closest thing to a definition of _mode_ would be
> The Upper and Lower code sets contain upper and lowercase alphabetic characters respectively; Mixed contains a mixture of ASCII control and punctuation characters; Punctuation contains common punctuation characters and combinations; and Digit contains the numerals plus some punctuation (..)<br>Message encoding starts in the Upper code set, and can be latched (semi-permanently, via U/L, L/L, M/L, P/L, or D/L) or shifted (for one character only, via U/S or P/S) into the other _modes_ as needed.

[...]

PR #643, opened 2016-08-02
Aztec decoder fix: return to Upper mode after U/S B/S sequence.

fgrieu:
Fix [issue #642](https://github.com/zxing/zxing/issues/642) (Aztec decoding: incorrect decoding of codes with U/S B/S).
This is a 1-line addition to core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
Add test case core/src/test/resources/blackbox/aztec-1/dlusbs.png and matching .txt
Change number of test cases.

The original code fails tests for core on the added test case; the source change fixes that.


srowen:
Great, thanks a lot for this.

PR #650, opened 2016-08-16
I've seen some codes with the type set to WPA2

steelman:
Make zxing recognize WIFI qrcodes with the type field set to WPA2 as WPA.


srowen:
Do you know if it actually works ? WPA isn't the same as WPA2. If the same APi supports it that is fine.

steelman:
I've decoded the string on my PC, removed "2", encoded it again and decoded in on a mobile which connected properly.

PR #652, opened 2016-08-21
Optimize QR data mask recognition

UniQP:


srowen:
Nice, yeah those are valid.

PR #653, opened 2016-08-22
Further micro-optimize QR data mask recognition; more comment for earlier such changes.

fgrieu:
Further micro-optimize QR data mask recognition; more comment for earlier such changes.


fgrieu:
I have experimentally confirmed that the the existing com.google.zxing.qrcode.decoder.DataMaskTestCase properly insures that the micro-optimization made to the DATA_MASK_110 code is correct.

srowen:
Yeah I reasoned through the cases and this is correct.

Issue #644, opened 2016-08-02
Core: DecoderResult's rawBytes comes without bit length indication, ambiguity arises


fgrieu:
The [DecoderResult class](https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/common/DecoderResult.java#L28) has a getRawBytes() method returning a byte array with the raw code's content. Problem is, there's no way to know exactly how many bits are in there: up to 7 bits in the last byte can be unused (in which case they are zeroes, and indistinguishable from actual bits at zero). This can be a cause of ambiguity, to the point that codes with different contents can have the same RawBytes, defeating its use as an inspection tool.

For examples, these two Aztec codes both have RawBytes<br>`f5 55 55 75 6b 5a d6 b5 ad 6b 5a d6 b5 ad 6b 5a d6 b5 ad 6b 5a d6 b0`<br>but the left one one contains<br>`88888TTTTTTTTTTTTTTTTTTTTTTTTTTTTTT`<br>while the right one contains<br>`88888TTTTTTTTTTTTTTTTTTTTTTTTTTTTTTO`<br>
![t](https://cloud.githubusercontent.com/assets/7809433/17342852/69115698-58fb-11e6-8e4d-846751baa59a.png) &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; ![to](https://cloud.githubusercontent.com/assets/7809433/17340857/46a7eada-58f2-11e6-8711-e7c5f4fed16b.png)
Both are plain vanilla Aztec codes generated with [bwip-js](http://bwip-js.metafloor.com/demo/demo.html) using `format=full eclevel=31` and their respective content. At the end of the rawBytes, the `b0` byte actually contains the bits `1001` in the left code and `10010000` in the right code, and that changes the decoded content.

The issue is one of design of the DecoderResult class. I see two ways to fix this:
1. Extend the DecoderResult class (in package com.google.zxing.common) with a getRawBytesBitSize() method returning an int telling how many bits there are in what getRawBytes() returns. This will leave any existing code working (including zxing's own test code), and requires no change in existing decoders when they never leave unused bits in RawBytes, or when that somewhat does not matter.
[...]

PR #664, opened 2016-09-02
Add 'getNumBits' to Result objects to return when not all bits of the raw byte representation are valid

srowen:
See https://github.com/zxing/zxing/issues/644
CC @fgrieu

### DIFF
--- a/core/src/main/java/com/google/zxing/Result.java
+++ b/core/src/main/java/com/google/zxing/Result.java
@@ -28,6 +28,7 @@ public final class Result {
 
   private final String text;
   private final byte[] rawBytes;
+  private final int numBits;
   private ResultPoint[] resultPoints;
   private final BarcodeFormat format;
   private Map<ResultMetadataType,Object> resultMetadata;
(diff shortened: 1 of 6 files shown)


fgrieu:
This resolves issue #644 as far as core is concerned. I wish getNumBits would be used, in particular by the web interface, but that requires changes to at least [decode.jspx](https://github.com/zxing/zxing/blob/master/zxingorg/src/main/webapp/w/decoderesult.jspx), and I'm incompetent for that.

Thank you for the good work! It was more involved than I initially thought, and making the right design choice was hard for me (opting for the additional-constructor pattern for zxing.Result, and the additional-setter pattern for zxing.common.DecoderResult, as you did).

srowen:
Yeah I punted on displaying this info for now. Maybe later it becomes clear how to do that cleanly.

dengm129:
https://github.com/zxing/zxing/commit/75d6000c1d87c0d179b3311b4c076c15dc527eb7

PR #666, opened 2016-09-09
translation contribution - issue on translate quality

Steinvikel:
This is my first action on GitHub... be lenient
(i speak only french & english... mostly french)

The French description of the application is syntactically correct, but semantically fairly consistent ... some passages still misunderstood at the time I speak.
I would like to translate the presentation text on the Google playstore in French (the English text into French), but I don't know where (or how) to find it in english.
I have looked for briefly, but i find no marks in name folders to get closer of this file.

Additionally, some informations that I think relevant are absent.
Should i only translate existing file ?.. or can i improve it with some details ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,19 @@
 language: java
 sudo: false
-install: mvn -T 2 -Ptravis -DskipTests=true -B -V install
-script: mvn -T 2 -Ptravis ${JACOCO} test
-matrix:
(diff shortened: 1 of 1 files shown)


Steinvikel:
(error in my pulling >> wrong "from" ... work in progress)

srowen:
OK, yeah looks like the wrong change.

PR #676, opened 2016-09-22
enterprise wifi support, requires android version 18

Grunge:
Hello There
Support for WPA2 Enterprise Wifi Networks.
Requires Android Version 18.
But xzing bases an earlier Version, so this does NOT compile!


srowen:
I'm not going to be able to merge this if it requires API 18, sorry. The app is still on API 15 in the current release, and API 16 for the next one sometime early next year. Setting aside a big legacy Android 2.x user base, there are still about 2.5 million users on API 15 alone, and I don't want to stop updating for them just yet, and also don't really want to maintain multiple branches.

PR #677, opened 2016-09-26
Specify ISO-8859-1 default encoding when guessing binary

l-dobrev:
When decoding QR Code in binary mode assume encoding ISO-8859-1 instead of the platform default.


srowen:
Ah right, I was going to suggest instead including the ECI segment when encoding even if it encodes ISO-8859-1. Because that will also always override the guessing here, I believe. And, you can force it with a decoding hint, too. What do you think of that instead?

The only odd thing about this change is that you return ISO-8859-1 knowingly when it's been 'ruled out'. But, in general the platform default might have been one of the ones ruled out anyway.

I'm reluctant to make this change because I suspect it ends up making some Chinese QR codes not work, because they (incorrectly) use a Chinese encoding, and hey at least this would often end up getting it right on Chinese devices.

PR #683, opened 2016-10-05
removed 'continuous-video' mode

verianel6869:
I found it better to not use this mode becouse it sets focus on start and then never tweak it again later. 

It is returned in AutofocusManager as currentFocusMode and maks useAutoFocus == false.

For me it worked better if currentFocusMode is 'auto' than 'continous-video' becouse it start autofocusing by focus() and autoFocusAgainLater() by AutoFocusManager. 

"... The focus position is locked after autoFocus call ...."
source:
https://developer.android.com/reference/android/hardware/Camera.Parameters.html#FOCUS_MODE_CONTINUOUS_VIDEO


srowen:
If useAutoFocus is false, then autoFocus() isn't called. The docs say it only locks if this is called. I think it's OK then?

verianel6869:
Ok, you are right

but on some older devices (Samsung GT-I9100) focus mode 'continuous-video' does not want to work.

I thought it gains focus at start and do not change it later.

srowen:
Hm, it seems like a device problem then. I hesitate to do this because it's a generally useful focus mode.

PR #695, opened 2016-10-26
resolve the problem of how to remove padding.

benjaminwhx:
My project need the QR code remove the padding, I try to set
`EncodeHintType.MARGIN = 0`
but not resolve my problem, so I add these codes to help people who have the problem with me.


srowen:
No, the margin is mandatory. I don't want to add an option to remove it as it becomes invalid.

benjaminwhx:
why becomes invalid?

PR #718, opened 2016-12-03
Change README.md. Added Delphi/Pascal ZXing reference.

Spelt:
Thanks for the great job and inspiration.

Sincerely,


Spelt:
Thanks!

2016-12-03 10:11 GMT+01:00 Sean Owen <notifications@github.com>:

PR #725, opened 2016-12-20
Async Open Driver call.

bhargavms:


srowen:
Interesting, but what does this help? the app can't do anything until the camera opens anyway, so it kind of blocks on this anyway. Other init is trivial.

bhargavms:
@srowen Android has a strict rule that you shouldn't perform long running executions on the main thread, this essentially follows that principle and offloads `camera.open()` to background worker thread, I can add more improvements on this where in I also offload `camera.release()` to this worker thread. Blocking the UI thread is a no no in my book, also the difference in User experience is that of between heaven and earth. You can run the apk on your device and have a look yourself. There is a 50-100 frames drop due to the `camera.open` call which is removed in this PR.

bhargavms:
@srowen any updates on this? Can you at least explain why you are adamant about not having this feature put in? I mean it doesn't beak anything, and it actually improves activity launch speed, I have tested this on more than 10 manufacturer devices (including samsung) and it works perfectly normal. May I know whats keeping you apprehensive?

srowen:
I don't understand what this helps, because the app must block (from the user's perspective) until the camera is ready no matter what. Yes, you could conceivably do other init while waiting, but it takes essentially no time. Blocking the user thread doesn't matter here because the app is not otherwise active.

More generally, I do not have any time to maintain the app anymore, and I can't test this or support it if it doesn't work. I think it's a bit risky as every time I've changed something with the camera in the past it's broken some device's odd driver behavior. The current behavior is, at least, quite safe.

bhargavms:
Well the reason I'm suggesting this change is because I've gotten complaints from users that the app stutters when opening the QrScanner (CaptureActivity). 
On old devices `Camera.open()` takes more than 100ms which creates a bad user experience. This change smooths it out.
Anyway like you said a change like this has to be backed up by lot of testing, the reason I gave the PR is in hopes of getting some help in testing and possibly fixing the qr scan speed which I'm not able to figure out.

PR #728, opened 2016-12-21
Integration of Privacy Friendly QR Scanner

Yonjuni:
Adding of Privacy Friendly QR Scanner (https://github.com/SecUSo/privacy-friendly-qr-scanner) to TARGET_ALL_KNOWN list. See https://github.com/SilenceIM/Silence/issues/330 

I hope this is done correctly ;-)

### DIFF
--- a/android-integration/src/main/java/com/google/zxing/integration/android/IntentIntegrator.java
+++ b/android-integration/src/main/java/com/google/zxing/integration/android/IntentIntegrator.java
@@ -120,6 +120,7 @@ public class IntentIntegrator {
 
   private static final String BS_PACKAGE = "com.google.zxing.client.android";
   private static final String BSPLUS_PACKAGE = "com.srowen.bs.android";
+  private static final String PFA_PACKAGE = "com.secuso.privacyFriendlyCodeScanner";
 
   // supported barcode formats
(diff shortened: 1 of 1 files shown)


srowen:
I'd prefer not to merge this because the list is intended to be only the app's package and 'official' derivatives. It's a bit of a slippery slope otherwise: how many more to list? how to verify they faithfully support the intent? the package name of the app you propose implies the main app isn't 'privacy safe', which isn't a bit true.

What we could however look at is allowing targeting no package at all and letting any app respond to the intent. The problem is potentially triggering an app that doesn't actually properly support the intent, but, suppose that is up to the caller since it wouldn't be the default.

Yonjuni:
Thanks for the answer. As Privacy Friendly QR Scanner belongs to the group of Privacy Friendly Apps (more info: secuso.org/pfa) it includes some mechanisms to enhance the user's privacy, but this does not mean, that it is the only privacy-friendly application and better than all others. And we furthermore don't assume that other apps have privacy issues by default.  
We have send the PR as a user request (see https://github.com/SilenceIM/Silence/issues/330).  I guess your suggestion could be a solution and of course it's up to you to decide.

PR #733, opened 2016-12-29
Added camera flip functionality.

JacobVanAssche:
User can now swap between Front and Rear facing camera on the screen. Preference will be saved.


srowen:
I think the problem is that this ignores the existing control over the camera selection, including dealing with real cases like not having a front camera. I would not be able to merge this.

Marylatha:
Porn sex hard core

Marylatha:
Undressed  sex

Marylatha:
Sex

PR #739, opened 2017-01-05
Add support for non-interleaved/standard two-of-five

l-dobrev:
This is a copy-and-paste-and-change of the Interleaved 2-of-5 reader to support the non-interleaved or standard variant of the Bar-Code.
Two black-box tests also supplied, test suit passes.


core/src/main/java/com/google/zxing/oned/STFReader.java
```diff
@@ -0,0 +1,352 @@
...
+
+  private static final int W = 3; // Pixel width of a wide line
+  private static final int S = 2; // Pixel width of a start/stop wide line
+  private static final int N = 1; // Pixel width of a narrow line
+
+  /** Valid ITF lengths. Anything longer than the largest value is also allowed. */
+  private static final int[] DEFAULT_ALLOWED_LENGTHS = { 6, 8, 10, 12, 14 };
+
+  // Stores the actual narrow line width of the image being decoded.
+  private int narrowLineWidth = -1;
```
srowen: It seems like this can easily not be a member, but a local var / argument?

Is there any more opportunity to refactor similarities with the ITF reader to reduce duplication?

Also, what's the general use case for this format? I hadn't heard of it.

I'm weighing usefuleness vs downside of copying and pasting code

l-dobrev:
I've attempted to change as little as possible.
The narrow line width is a side-result of checking for the start pattern (it's calculated there). I suppose a two phase process could be implemented.
I've attempted to re-factor the code so that both standard and interleaved 2-of-5 are read with the same class. This came out to be very messy, since the two codes are similar, but the differences are quite significant. I have not tried re-factoring a common abstract base, but it seems achievable.
This symbology is somewhat side-lined, since a better Interleaved variant has outscored it, but the issue has been brought more than once on the mailing list. 
This symbology seems to be the Industrial or Standard 2-of-5. The only source about this code I can find is http://www.barcodeisland.com/int2of5.phtml.
Unlike the Interleaved variant this symbology can be written using a font: http://www.barcodefonts.net/Industrial2of5_Barcode.html
The symbology is only mentioned on the Wikipedia https://en.wikipedia.org/wiki/Barcode#Linear_barcodes

PR #756, opened 2017-02-18
Correct size of array in ITFWriter

dsanders11:
The intermediate array only needs to be 10 ints long, not 18. The max value for `j` is 4, so `2 * 4 + 1 = 9`.


srowen:
Doesn't it need to be 10 if the max index is 9?

dsanders11:
@srowen, yes, you're correct, whoops. I'm currently porting the 1D barcode code to JavaScript as part of extending [Schibum/barcode.js](https://github.com/Schibum/barcode.js/) to also work with 1D barcodes which is why I noticed this. Having the array be too large was causing a failure because of a for loop over the patterns running into uninitialized values (NaN). Since arrays in JavaScript are dynamic by making it 9 instead of 10 it still worked because the array just expanded to 10 and I didn't notice. Too many late nights, not enough coffee.

I'll fix pull request to change it to 10.

PR #780, opened 2017-03-31
Update README.md

rbashish:
add an CI Status


srowen:
This is already at the bottom with other badges.

rbashish:
Oh yes, I realized that. But don't you think it's better to keep on top? Any way decision is yours.

PR #793, opened 2017-04-24
Update ZXing.NET project link

KrasnayaPloshchad:
The ZXing.NET project is originally hosted on CodePlex, but Microsoft shut down the site recently, then the developer moved it to GitHub. So I update the link.


KrasnayaPloshchad:
🤝

PR #795, opened 2017-04-29
findBestPreviewSizeValue: sorting optimized

a-ch:
It's more effective to perform sorting on the supportedPreviewSizes list with unsupported values previously removed

### DIFF
--- a/android-core/src/main/java/com/google/zxing/client/android/camera/CameraConfigurationUtils.java
+++ b/android-core/src/main/java/com/google/zxing/client/android/camera/CameraConfigurationUtils.java
@@ -284,23 +284,7 @@ public final class CameraConfigurationUtils {
       return new Point(defaultSize.width, defaultSize.height);
     }
 
-    // Sort by size, descending
     List<Camera.Size> supportedPreviewSizes = new ArrayList<>(rawSupportedSizes);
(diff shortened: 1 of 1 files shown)


srowen:
Yeah you don't even need to sort then. At that point all that matters is the one with max resolution. 
I don't think the sort is expensive and makes the log message that follows a little more orderly. 
That said, maybe in some corner case it is better to take an acceptable option that was presented earlier by the device. 
If you'll take it the extra logical step yes that's ok with me.

srowen:
@a-ch if you'll update accordingly I'll merge it

a-ch:
Sorry, I don't really get what exactly you're asking me to update in the code. Do you mean that it would be better to remove sorting at all and simply find the max resolution?

srowen:
Yes, because at the point the sort happens now, afterwards, only the first element is used. Therefore you can remove the sort entirely and just find the max resolution.

srowen:
Or, I think you don't even need to copy the collection of values and then prune it. You can just add the suitable sizes to the collection. And then, you don't even need to do that -- just track the suitable size with max resolution. I have a PR if you want to take a look

srowen:
Pardon, I addressed this directly in https://github.com/zxing/zxing/commit/3d0468e92f0cc8a7fba9c6e1355e3f7bc1f7d93f

PR #800, opened 2017-05-10
Clean up whitespace

dsanders11:
Just some code hygiene cleanup. Dropped trailing or excess whitespace, and also dropped trailing commas in array initializers, as the dominant style in the codebase is to not have them and that fits more with other languages.


core/src/main/java/com/google/zxing/maxicode/decoder/DecodedBitStreamParser.java
```diff
@@ -144,7 +144,7 @@ private static String getPostCode3(byte[] bytes) {
          SETS[0].charAt(getInt(bytes, new byte[] {27, 28, 29, 30, 19, 20})),
          SETS[0].charAt(getInt(bytes, new byte[] {21, 22, 23, 24, 13, 14})),
          SETS[0].charAt(getInt(bytes, new byte[] {15, 16, 17, 18,  7,  8})),
-         SETS[0].charAt(getInt(bytes, new byte[] { 9, 10, 11, 12,  1,  2})),
+         SETS[0].charAt(getInt(bytes, new byte[] { 9, 10, 11, 12,  1,  2}))
```
srowen: I leave things like this in on purpose. It's really trivial, but if you add a new line, at least you don't have to change this line to add the comma. The symmetry is nice.

dsanders11: I definitely see the benefits of the trailing commas, but I'll make two arguments against. Wall of text to follow, but I don't have a strong opinion either way.

First, most of the code does not have a trailing comma, so it is inconsistent in the codebase. There's only a few in this commit that I found, which shows that the majority of them don't have a trailing comma. Regardless of preferred style it would be nice if it was consistent.

Second, other languages often don't allow trailing commas. This may seem trivial, but I'm currently porting as much of ZXing as I can to JavaScript, and the two languages are close enough that sometimes it's just a matter of tweaking some syntax and keywords. Trailing commas in arrays aren't allowed in JavaScript, though, so I hit that a few times because it's easy to not see visually when porting hundreds of lines. So I'd make the argument that it makes the code ever so slightly more accessible for porting to other languages without the trailing commas. Obviously language-specific features shouldn't be avoided for this reason, but in this case its something trivial.

Anyway, not a strong opinion, but there's my rationale. I can remove the comma relate edits if you'd still like to keep the trailing commas.

dsanders11: It appears trailing commas are now allowed in JavaScript, not sure what I was running into, too many late nights, may have been seeing things. My arguments still stand, but I've gone ahead and reverted the trailing comma edits so now the PR is just whitespace.

kitsuphat:
Today is the action. You will see it here.

PR #811, opened 2017-06-08
Fixed Wifi mangager memory leak on devices < N

PhilLab:
AndroidStudio reported the error:
Error: The WIFI_SERVICE must be looked up on the Application context or memory will leak on devices < Android N. Try changing to .getApplicationContext()  [WifiManagerLeak]


srowen:
@PhilLab OK, is there any other implication to looking up from context? 
What about other services that are looked up this way?

PhilLab:
@srowen It seems that this is a specific issue for the WIFI_SERVICE. I checked a couple of other ``getSystemService()`` calls and none reported this issue. A [google search](https://www.google.de/search?q=must%20be%20looked%20up%20on%20the%20Application%20context%20or%20memory%20%20will%20leak%20on%20devices%20<%20Android%20N.) only brought up the wifi_service but this is of course only an indication, not a proof

Issue #126, opened 2014-04-28
Support for WPA-Enterprise?


ISharkI:
Hello

We just stumbled upon the problem that we want to grand our members (4k+) access to our distributed wifi network which uses 802.1X for authentication.
Currently the scanner supports only "open","WEP", "WPA" and "WPA2".

According to my current understanding of the code (if I missed something please tell me) there are only the files 

core/src/main/java/com/google/zxing/client/result/WifiResultParser.java
core/src/main/java/com/google/zxing/client/result/WifiParsedResult.java

which generate the config and is parsed to the Android manager later. According to the Android documentation the WifiConfiguration accepts "public WifiEnterpriseConfig enterpriseConfig" with all neccessary sub-options like "public int getPhase2Method ()" etc. However testing in our network (with it's current configuration) using an QR-Code was not successful. 

Am I wrong? If yes I would like to know where my mistake is.
If you need further information in order to reproduce the issue don't hesitate to contact me.

Greetings
Michael

denysvitali:
Sorry for the necro-bump, but it looks like 802.11x is still not supported. Is there any plan to do it?

srowen:
I'd be OK with requiring Android 4.3+ at this point. So if you want to revive this and propose a PR @denysvitali I'd look at it.

PR #865, opened 2017-09-03
Implement support for WPA2 Enterprise (EAP)

steffen-kiess:
This adds support for parsing QR codes which contain information about an WPA2 enterprise (EAP) network.

For WPA2 enterprise, strings will be of the form:
WIFI:T:WPA2-EAP;S:[network SSID];H:[hidden?];E:[EAP method];PH2:[Phase 2 method];AI:[anonymous identity];I:[username];P:[password];;
The "EAP method" can e.g. be "TTLS" or "PWD" or one of the other fields in [WifiEnterpriseConfig.Eap ](https://developer.android.com/reference/android/net/wifi/WifiEnterpriseConfig.Eap.html) and "Phase 2 method" can e.g. be "MSCHAPV2" or any of the other fields in [WifiEnterpriseConfig.Phase2](https://developer.android.com/reference/android/net/wifi/WifiEnterpriseConfig.Phase2.html)

An example would be:

    WIFI:S:eduroam;T:WPA2-EAP;AI:anonymous@uni-stuttgart.de;I:username@uni-stuttgart.de;P:password;E:PEAP;PH2:MSCHAPV2;;

### DIFF
--- a/android/src/com/google/zxing/client/android/wifi/WifiConfigManager.java
+++ b/android/src/com/google/zxing/client/android/wifi/WifiConfigManager.java
@@ -17,6 +17,7 @@
 package com.google.zxing.client.android.wifi;
 
 import android.net.wifi.WifiConfiguration;
+import android.net.wifi.WifiEnterpriseConfig;
 import android.net.wifi.WifiManager;
 import android.os.AsyncTask;
 import android.util.Log;
@@ -28,6 +29,7 @@ import com.google.zxing.client.result.WifiParsedResult;
 /**
  * @author Vikram Aggarwal
  * @author Sean Owen
+ * @author Steffen Kieß
  */
 public final class WifiConfigManager extends AsyncTask<WifiParsedResult,Object,Object> {
 
@@ -88,6 +90,9 @@ public final class WifiConfigManager extends AsyncTask<WifiParsedResult,Object,O
             break;
           case WPA:
             changeNetworkWPA(wifiManager, theWifiResult);
+            break;
+          case WPA2_EAP:
+            changeNetworkWPA2EAP(wifiManager, theWifiResult);
             break;
         }
       }
@@ -164,6 +169,26 @@ public final class WifiConfigManager extends AsyncTask<WifiParsedResult,Object,O
     updateNetwork(wifiManager, config);
   }
 
+  // Adding a WPA2 enterprise (EAP) network
+  private static void changeNetworkWPA2EAP(WifiManager wifiManager, WifiParsedResult wifiResult) {
+    WifiConfiguration config = changeNetworkCommon(wifiResult);
+    // Hex passwords that are 64 bits long are not to be quoted.
+    config.preSharedKey = quoteNonHex(wifiResult.getPassword(), 64);
+    config.allowedAuthAlgorithms.set(WifiConfiguration.AuthAlgorithm.OPEN);
+    config.allowedProtocols.set(WifiConfiguration.Protocol.RSN); // For WPA2
+    config.allowedKeyManagement.set(WifiConfiguration.KeyMgmt.WPA_EAP);
+    config.allowedPairwiseCiphers.set(WifiConfiguration.PairwiseCipher.TKIP);
+    config.allowedPairwiseCiphers.set(WifiConfiguration.PairwiseCipher.CCMP);
+    config.allowedGroupCiphers.set(WifiConfiguration.GroupCipher.TKIP);
+    config.allowedGroupCiphers.set(WifiConfiguration.GroupCipher.CCMP);
+    config.enterpriseConfig.setIdentity(wifiResult.getIdentity());
+    config.enterpriseConfig.setAnonymousIdentity(wifiResult.getAnonymousIdentity());
+    config.enterpriseConfig.setPassword(wifiResult.getPassword());
+    config.enterpriseConfig.setEapMethod(parseEap(wifiResult.getEapMethod()));
+    config.enterpriseConfig.setPhase2Method(parsePhase2(wifiResult.getPhase2Method()));
+    updateNetwork(wifiManager, config);
+  }
+
   // Adding an open, unsecured network
   private static void changeNetworkUnEncrypted(WifiManager wifiManager, WifiParsedResult wifiResult) {
     WifiConfiguration config = changeNetworkCommon(wifiResult);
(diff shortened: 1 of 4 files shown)


steffen-kiess:
This was also discussed in #126

android/src/com/google/zxing/client/android/wifi/WifiConfigManager.java
```diff
@@ -17,6 +17,7 @@
 package com.google.zxing.client.android.wifi;
 
 import android.net.wifi.WifiConfiguration;
+import android.net.wifi.WifiEnterpriseConfig;
```
srowen: I think the reason this had been rejected before is that it requires API 18. I think that's OK to require at this point. But it would require updating the Manifest and build files to reflect that minimum too.

core/src/main/java/com/google/zxing/client/result/WifiResultParser.java
```diff
@@ -36,16 +43,21 @@ public WifiParsedResult parse(Result result) {
     if (!rawText.startsWith("WIFI:")) {
       return null;
     }
-    String ssid = matchSinglePrefixedField("S:", rawText, ';', false);
+    rawText = ";" + rawText.substring("WIFI:".length());
```
srowen: Why the semicolon in this and later lines?

core/src/main/java/com/google/zxing/client/result/WifiParsedResult.java
```diff
@@ -27,17 +27,29 @@
   private final String networkEncryption;
   private final String password;
   private final boolean hidden;
+  private final String identity;
+  private final String anonymousIdentity;
+  private final String eapMethod;
+  private final String phase2Method;
```
srowen: I was going to say we should probably break down this class rather than loading all these different properties into one place, but at this point, probably worth just tacking it on.

core/src/main/java/com/google/zxing/client/result/WifiResultParser.java
steffen-kiess: Otherwise there it will also match if one field name is a suffix of another, e.g. if you have

    WIFI:AI:anonidentity;I:identity;...;;

it will set both identity and anonymousIdentity to "anonidentity" (because it only searches for "I:"). I think the best way to avoid that is to require the field name to be directly after a ";" (or to be the first character after the "WIFI:".

android/src/com/google/zxing/client/android/wifi/WifiConfigManager.java
steffen-kiess: So I should update https://github.com/zxing/zxing/blob/87f00d41a747382a1dc34a6e67bfb3b21c4c4236/android/AndroidManifest.xml#L34 this line to say "android:minSdkVersion="18""?

core/src/main/java/com/google/zxing/client/result/WifiResultParser.java
srowen: OK, how about choosing different one-letter prefixes then, like A and H instead of AI and PH2? I know it's kind of cryptic but this is a very short simple format.

android/src/com/google/zxing/client/android/wifi/WifiConfigManager.java
```diff
@@ -226,4 +251,30 @@ private static boolean isHexOfLength(CharSequence value, int... allowedLengths)
     return false;
   }
 
+  private static int parseEap(String eapString) {
+    if (eapString == null) {
+      return WifiEnterpriseConfig.Eap.NONE;
+    }
+    try {
+      return (Integer) WifiEnterpriseConfig.Eap.class.getField(eapString).get(null);
```
srowen: Why was the reflection needed here? is it due to API levels?

core/src/main/java/com/google/zxing/client/result/WifiResultParser.java
steffen-kiess: I don't have any strong opinion on the name of the prefixes, so I'll change it to 'A' and 'H'.

Should the code recognize the prefixes only after a ';' (like it is currently in my pull request) or anywhere (like it is currently in zxing)?

android/src/com/google/zxing/client/android/wifi/WifiConfigManager.java
steffen-kiess: The advantage of reflection would be that zxing will support new methods without any changes. If you want, I can change this to an explicit list of methods without reflection, as all the important ones (PEAP, PWD, TLS, TTLS for Eap and MSCHAP, MSCHAPv2, PAP for Phase2) are available with API level 18.

steffen-kiess:
I've updated the pull request
- rebase against current master
- Use A: and H: as prefix for anonymous identity and phase2 method
- Recognize prefixes everywhere, not just after a ";"
- Don't use reflection, instead have a switch{} block with the methods available in API level 19

somoza:
Not working in the new Android versions? If I dont set "No Validate" in the certificate CA menu, I cannot create the connection...

srowen:
Not sure, I don't have any way to test it. You may be right, or it may be specific to the network you're joining.

somoza:
I checked it in Android Oreo, but I will test it in below versions...

somoza:
I checked in Android 6, same error... Is this sintax correct? Somebody had success with this?

WIFI:T:WPA2-EAP;S:SIGNALNAME;E:PEAP;H:MSCHAPV2;I:julian.somoza;P:PaS$;;

steffen-kiess:
If you have root access to your phone, maybe you can check what's the difference between adding the entry manually and adding it with zxing. The information about the wifi networks is on `/data/misc/wifi/wpa_supplicant.conf` (at least on my phone). There should be some block for the network you've added, like this:

    network={
        ssid="somewhat"
        bssid=12:34:56:78:90:ab
        proto=RSN
        ...
    }

If you can tell us what's the difference between manually adding it (in which case it works as far as I understand you) and adding it with zxing we should know what's missing.

czare:
I tried almost all strings from this thread in Android 10. All I seem to get is an entry for a WPA Personal network.
I couldn't get a WPA Enterprise one. Is this still working?

gurkburk76:
Can't get it to work, only got android 12 to test on non-rooted.
If anyone could post a know working config for it i'de be most greatfull.

paulopera:
Did anyone have any luck with iOS? The best site which I could find somethings was this https://jasonmurray.org/posts/2020/wifiqrcode/

The only way it kinda of "worked", was when I used [] and another : between the parameters.
something like this:

```
qrencode -o wifi.png "WIFI:S:[SSID-NAME];:T:[WPA2-EAP];:I:[USER];:P:[PASSWORD]:PH2:[MSCHAPV2];:I:[USER];:P:[PASSWORD];;"

```
The camera app shows **JOIN WIFI SSID-NAME**  and then click try to connect, if I remove the extra : and [] it shows a error that there's nothing in the qrcode.

abuziad2030:
How i can use it

LaneaLucy:
anything new? is it working?

PR #872, opened 2017-09-14
asterisk isn't a valid character and should not be encoded as content with Code 39.

micjahn:
the following sample content results in an unreadable barcode: *123456789
because the asterisk is the start-stop character (special meaning, not valid in content)
the really encoded value is **123456789*. The barcode reader stops after reading the second asterisk
and throw the content away.


core/src/main/java/com/google/zxing/oned/Code39Reader.java
```diff
@@ -49,11 +47,11 @@
       0x034, 0x121, 0x061, 0x160, 0x031, 0x130, 0x070, 0x025, 0x124, 0x064, // 0-9
       0x109, 0x049, 0x148, 0x019, 0x118, 0x058, 0x00D, 0x10C, 0x04C, 0x01C, // A-J
       0x103, 0x043, 0x142, 0x013, 0x112, 0x052, 0x007, 0x106, 0x046, 0x016, // K-T
-      0x181, 0x0C1, 0x1C0, 0x091, 0x190, 0x0D0, 0x085, 0x184, 0x0C4, 0x094, // U-*
-      0x0A8, 0x0A2, 0x08A, 0x02A // $-%
+      0x181, 0x0C1, 0x1C0, 0x091, 0x190, 0x0D0, 0x085, 0x184, 0x0C4, 0x0A8, // U-
```
srowen: Should the comment be "U-$" now? and the following line "/-%"?

core/src/main/java/com/google/zxing/oned/Code39Reader.java
```diff
@@ -262,6 +260,9 @@ private static char patternToChar(int pattern) throws NotFoundException {
         return ALPHABET_STRING.charAt(i);
       }
     }
+    if (pattern == ASTERISK_ENCODING) {
+        return '*';
```
srowen: Nit: 2-space indent

core/src/main/java/com/google/zxing/oned/Code39Reader.java
micjahn: yes. I tought it is the space character. was a mistake

PR #873, opened 2017-09-14
Fix extended mode decoding code39

micjahn:
The codes with preceding % are not linear arranged in the symbol table.
Changes are based upon https://en.wikipedia.org/wiki/Code39#Full_ASCII_Code_39


core/src/main/java/com/google/zxing/oned/Code39Reader.java
```diff
@@ -294,8 +294,20 @@ private static String decodeExtended(CharSequence encoded) throws FormatExceptio
...
             if (next >= 'A' && next <= 'E') {
               decodedChar = (char) (next - 38);
-            } else if (next >= 'F' && next <= 'W') {
+            } else if (next >= 'F' && next <= 'J') {
               decodedChar = (char) (next - 11);
+            } else if (next >= 'K' && next <= 'O') {
+              decodedChar = (char) (next + 16);
+            } else if (next >= 'P' && next <= 'T') {
+              decodedChar = (char) (next + 43);
+            } else if (next == 'U') {
```
srowen: Huh, so this never worked properly. 
These last 6 cases could be a switch but not sure if that's clearer or faster.
Do you have any test case handy that would test extended logic, so that it stays fixed?

micjahn: here we go... hopefully it is what you expected

PR #876, opened 2017-09-17
Code39 extended mode encoding support

micjahn:


core/src/main/java/com/google/zxing/oned/Code39Writer.java
```diff
@@ -50,13 +50,23 @@ public BitMatrix encode(String contents,
...
-    int codeWidth = 24 + 1 + length;
     for (int i = 0; i < length; i++) {
       int indexInString = Code39Reader.ALPHABET_STRING.indexOf(contents.charAt(i));
       if (indexInString < 0) {
-        throw new IllegalArgumentException("Bad contents: " + contents);
+        contents = tryToConvertToExtendedMode(contents);
+        length = contents.length();
+        if (length > 80) {
+          throw new IllegalArgumentException(
+              "Requested contents should be less than 80 digits long, but got " + length + " (extended full ascii mode)");
```
srowen: Just some nits here. ascii -> ASCII

core/src/main/java/com/google/zxing/oned/Code39Writer.java
```diff
@@ -86,4 +96,74 @@ private static void toIntArray(int a, int[] toReturn) {
     }
   }
 
+  private static String tryToConvertToExtendedMode(String contents) throws IllegalArgumentException {
```
srowen: You can omit IllegalArgumentException; it's not checked

core/src/main/java/com/google/zxing/oned/Code39Writer.java
```diff
@@ -86,4 +96,74 @@ private static void toIntArray(int a, int[] toReturn) {
...
+     int length = contents.length();
+     StringBuilder extendedContent = new StringBuilder();
+     for (int i = 0; i < length; i++) {
+       int character = (int)contents.charAt(i);
+       switch (character) {
+         case 0:
+           extendedContent.append("%U");
+           break;
+         case 32:
+           extendedContent.append(" ");
```
srowen: In these cases, append a char instead of a 1-char string

core/src/main/java/com/google/zxing/oned/Code39Writer.java
```diff
@@ -86,4 +96,74 @@ private static void toIntArray(int a, int[] toReturn) {
     }
   }
 
+  private static String tryToConvertToExtendedMode(String contents) throws IllegalArgumentException {
+     int length = contents.length();
+     StringBuilder extendedContent = new StringBuilder();
+     for (int i = 0; i < length; i++) {
+       int character = (int)contents.charAt(i);
```
srowen: Nit: space after cast. But can you not switch on a char? It might be more descriptive because the case labels would be chars

core/src/main/java/com/google/zxing/oned/Code39Writer.java
```diff
@@ -86,4 +96,74 @@ private static void toIntArray(int a, int[] toReturn) {
...
+         case 96:
+           extendedContent.append("%W");
+           break;
+         default:
+           if (character > 0 &&
+               character < 27) {
+             extendedContent.append("$");
+             extendedContent.append((char)('A' + (character - 1)));
+           }
+           else if (character > 26 && character < 32) {
```
srowen: Pull these up on the previous line

core/src/main/java/com/google/zxing/oned/Code39Writer.java
```diff
@@ -86,4 +96,74 @@ private static void toIntArray(int a, int[] toReturn) {
...
+           break;
+         case 64:
+           extendedContent.append("%V");
+           break;
+         case 96:
+           extendedContent.append("%W");
+           break;
+         default:
+           if (character > 0 &&
+               character < 27) {
```
srowen: This should be on one line

core/src/main/java/com/google/zxing/oned/Code39Writer.java
```diff
@@ -86,4 +96,60 @@ private static void toIntArray(int a, int[] toReturn) {
     }
   }
 
+  private static String tryToConvertToExtendedMode(String contents) {
+     int length = contents.length();
+     StringBuilder extendedContent = new StringBuilder();
+     for (int i = 0; i < length; i++) {
+       char character = contents.charAt(i);
+       switch (character) {
+         case '\u0000':
```
srowen: I think this can be '\0' for consistency but don't change unless something else needs to change

micjahn:
of course, every content which was invalid for code 39 without full ASCII support is now valid and the code 39 barcode is generated automatically in extended mode. but in my opinion this is a desired behavior. on the other hand every content which is invalid in both cases stays invalid.

PR #887, opened 2017-10-12
Fixed error key in doPost()

xuthus:
doPost() works only with badimage


srowen:
That's fine. Really we could use more different error messages maybe but that's more appropriate for an I/O error yes

PR #892, opened 2017-10-15
Update Google Play Badge

SabineJuergensen:
[https://play.google.com/intl/en_us/badges/](https://play.google.com/intl/en_us/badges/)


srowen:
I'm OK with that. You have a copy of the image here in order to make it smaller? I think we can also just link to the original and change the size?

```
[<img height='62' width='161' src='https://play.google.com/intl/en_us/badges/images/generic/en_badge_web_generic.png'/>](https://play.google.com/store/apps/details?id=com.google.zxing.client.android)
```

srowen:
I went ahead and merged a variant on this.

PR #898, opened 2017-10-28
Creating a values-nb folder for Norwegian Bokmål

DandelionSprout:
I couldn't see the new Norwegian translation in Barcode Scanner 4.7.7 on either of my two Android 7.1.x phones, and while the following is hard to confirm due to the lack of any hard info about it, I now feel convinced that the values-no tag must've been deprecated in recent Android versions. Therefore I am creating an identical strings.xml file in a values-nb folder.

I pondered on whether I should delete the values-no folder as well, but it was so hard to find info on the matter, that for all I know it could still be necessary for some Android 4 units or something like that.

### DIFF
--- a/android/res/values-nb/strings.xml
+++ b/android/res/values-nb/strings.xml
@@ -1,0 +1,140 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright (C) 2014 ZXing authors
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


srowen:
What's the difference between 'no' and 'nb'? how important is it to maintain both?

DandelionSprout:
As far as the ISO codes' definitions are concerned, 'no' is for the entire Norwegian language, while 'nb' is for the Bokmål written variant, which is used by around 90% of all Norwegians.

As far as Android is concerned, it's a big mess, in which some phones and/or Android versions only recognise the 'no' tag, while others only recognise the 'nb' tag; and I am semi-qualifiedly assuming that 'nb' has become the default in recent Android versions.

I have no idea if there are any phones, especially from Android 5 onwards, that are still using the values-no tag in any meaningful way. If in doubt, I'd advocate deleting the values-no folder for simplicity's sake, however.

srowen:
Is it valid enough to have the same strings for both? if so, this can just be as symlink from values-nb -> values-no

DandelionSprout:
That sounds to me like an excellent solution, if I understand it right. I'd say go for it.

srowen:
Done in `master`.

Issue #897, opened 2017-10-27
Finding false QR Code centers


herman1vdb:
Problematic QR Code:
![image](https://user-images.githubusercontent.com/8955362/32093202-cb1d737c-bafb-11e7-899e-98686c6d4b55.png)

Finding (also changing the angle of the QR code to 45° scans successfully):
![image](https://user-images.githubusercontent.com/8955362/32093249-f5e1f402-bafb-11e7-84c1-daaf8513821a.png)

After further investigation it seems the vertical and horizontal pattern is accepted at the above incorrect center, for this particular QR Code.
Which lead me to come across these concerns:

It seems at [FinderPatternFinder.java:78](https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java#L78)
`pureBarcode = hints != null && hints.containsKey(DecodeHintType.PURE_BARCODE);`
the default for `pureBarcode` is false (I'd rather assume most QR Codes are pure monochrome?)
eg. `pureBarcode = hints == null || hints.containsKey(DecodeHintType.PURE_BARCODE);`

In any case at [FinderPatternFinder.java:496](https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java#L496) it seems that the diagonal check is ignored when it is not monochrome (`!pureBarcode || crossCheckDiagonal(...`). This seems strange can someone please explain the reasoning behind this, I don't understand it. According to me it should still do the check regardless if it is a `pureBarcode` or not.

I'll submit a pull request if required.

**meaning of Pure Barcode:**
```
  /**
   * Image is a pure monochrome image of a barcode. Doesn't matter what it maps to;
   * use {@link Boolean#TRUE}.
   */
  PURE_BARCODE(Void.class),
```
_DecodeHintType.java_

srowen:
Yes, exactly, the heuristic will accept those because they look like finder patterns if you only look above and to the sides. It's optimizing for images of barcodes, where this false positive essentially doesn't happen.

"Pure barcode" mode was made for these cases, because frequently people want to decode a barcode image like these, not a picture, and should be trivial to find the corners just by looking at where the black pixels start. It is not the default, no. It is not a mode for black-and-white in general. It is a mode for non-photos.

This works with PURE_BARCODE on, right? does for me.

[...]

PR #906, opened 2017-11-07
Add QRCode lenient diagonal check

herman1vdb:
Adding diagonal check to solve #897 to prevent false positives for centers.


srowen:
OK, I buy it. I can see from the logs that it improves the tests by 6 images overall -- see the messages with 'lax' in them. Are you able to update the tests to expect the increased number of successes?

herman1vdb:
Cool, I've updated the tests.

PR #926, opened 2017-12-18
Add support for encoding GS1 formatted QR

dpeckett:
This PR adds support for encoding GS1 formatted QR's through the addition of an encoding hint. Basically for the QR to meet the GS1 standard we must prefix the header bit stream with the FNC1 mode header. Prior to this PR there was no means of adding this header.


dpeckett:
Closed in order to improve test coverage in a separate branch.

PR #927, opened 2017-12-18
Add support for encoding GS1 formatted QR

dpeckett:
This PR adds support for encoding GS1 formatted QR's through the addition of an encoding hint. Basically for the QR to meet the GS1 standard we must prefix the header bit stream with the FNC1 mode header. Prior to this PR there was no means of adding this header.

### DIFF
--- a/core/src/main/java/com/google/zxing/EncodeHintType.java
+++ b/core/src/main/java/com/google/zxing/EncodeHintType.java
@@ -101,4 +101,10 @@ public enum EncodeHintType {
     * (Type {@link Integer}, or {@link String} representation of the integer value).
     */
    QR_VERSION,
+
+  /**
+   * Specifies whether the data should be encoded to the GS1 standard (type {@link Boolean}, or "true" or "false"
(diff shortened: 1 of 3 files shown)


srowen:
OK by me. Could you point me to any other info on how this differs? does it really require an explicit flag from the user? probably so just wondering if there's anything simpler that can be done.

dpeckett:
Relevant excerpt from ISO/IEC 180004:2006 regarding the FNC1 header:

This  mode  indicator  identifies  symbols  encoding  data  formatted  according  to  the  GS1  Application  Identifiers standard. For this purpose, it shall only be used once in a symbol and shall be placed immediately before the first mode indicator used for efficient data encoding (Numeric, Alphanumeric, Byte or Kanji), and after any ECI or Structured Append header.

We don't currently have a means for setting this mode header, I don't think there's anything simpler available than a switch. 

The above excerpt has shown there's a bug with my PR, we want the FNC1 mode header before the encoding but after any ECI headers. I will submit a revision for this.

Issue #836, opened 2017-07-11
CommandLineEncoder appears to ignore the height and width parameters when barcode_format is  DATA_MATRIX


ablegreen:
`CommandLineEncoder` appears to ignore the height and width parameters when `barcode_format` is  `DATA_MATRIX`

The following command produces a 14px by 14px image instead of a 500px by 500px image: 

```
java -cp javase/target/javase-3.3.0-jar-with-dependencies.jar com.google.zxing.client.j2se.CommandLineEncoder --barcode_format DATA_MATRIX --height 500 --width 500 abcdefg
```

Here's a `Dockerfile` to reproduce this issue:

```
FROM ubuntu:16.04
RUN apt-get update && apt-get install -y --no-install-recommends default-jdk git maven
RUN git clone --depth 1 --branch zxing-3.3.0 https://github.com/zxing/zxing.git
RUN cd zxing && mvn install -DskipTests
RUN cd zxing/javase && mvn -DskipTests package assembly:single
RUN cd zxing && java -cp javase/target/javase-3.3.0-jar-with-dependencies.jar com.google.zxing.client.j2se.CommandLineEncoder --barcode_format DATA_MATRIX --height 500 --width 500 abcdefg
```

Is this a bug or am I using `CommandLineEncoder` wrong?

srowen:
You're right, I don't believe it actually respects those values. The DataMatrix encoder was adopted from another project and works somewhat differently. If you're interested in looking at plumbing through those values and trying to make it scale up the image and add padding like the QR code writer does I can look at that with you. It's a legitimate issue, yes.

axxel:
Just as a reference for whom it may concern: I recently introduced this missing feature to the new c++ port with this change: https://github.com/nu-book/zxing-cpp/commit/be46af7ddc0da3b955d8547c19d81140a4c707b2

The interesting part is the new generic `Inflate` function that is now used inside the QRWriter, the AztecWriter and the DatamatrixWriter. It takes a `BitMatrix` and inflates it by scaling it up and adding padding as necessary.

iphone6fromApple:
i  on concern

iphone6fromApple:
i on concern

Medusa500:
DATA_MATRIX

wvieed:
V

janeneb711:
I know nothing about source code my last device that was breach is android and now I have an Apple iPhone trying to fix my name and recover hijacked gmail account

gredler:
[...]

PR #933, opened 2017-12-27
Proposed fix for Issue #836 - height and width parameters ignored when barcode_format is DATA_MATRIX

kent92:
Hi @srowen , was wondering if you'd like to go over my proposed fix for this issue. Tried it locally and it works, passes unit tests, etc. Let me know if further changes are needed and i will follow through accordingly. 

Thanks!

### DIFF
--- a/core/src/main/java/com/google/zxing/datamatrix/DataMatrixWriter.java
+++ b/core/src/main/java/com/google/zxing/datamatrix/DataMatrixWriter.java
@@ -38,7 +38,6 @@ import java.util.Map;
  */
 public final class DataMatrixWriter implements Writer {
 
-
   @Override
   public BitMatrix encode(String contents, BarcodeFormat format, int width, int height) {
     return encode(contents, format, width, height, null);
@@ -90,8 +89,7 @@ public final class DataMatrixWriter implements Writer {
     String codewords = ErrorCorrection.encodeECC200(encoded, symbolInfo);
 
     //3. step: Module placement in Matrix
-    DefaultPlacement placement =
-            new DefaultPlacement(codewords, symbolInfo.getSymbolDataWidth(), symbolInfo.getSymbolDataHeight());
+    DefaultPlacement placement = new DefaultPlacement(codewords, symbolInfo.getSymbolDataWidth(), symbolInfo.getSymbolDataHeight());
     placement.place();
 
     //4. step: low-level encoding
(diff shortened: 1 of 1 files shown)


axxel:
My 2 cents as an outsider: why not refactor the QRCode + AZtech + DataMatrix writers so they all use the same simple scaling/inflating code instead of adding yet another copy of the functionality? (I'm biased, of course: https://github.com/zxing/zxing/issues/836#issuecomment-345980338 ;))

core/src/main/java/com/google/zxing/datamatrix/DataMatrixWriter.java
```diff
@@ -38,6 +38,7 @@
  */
 public final class DataMatrixWriter implements Writer {
 
+
```
srowen: If you would, revert the extra space changes like this one and one below

core/src/main/java/com/google/zxing/datamatrix/DataMatrixWriter.java
```diff
@@ -159,17 +160,33 @@ private static BitMatrix encodeLowLevel(DefaultPlacement placement, SymbolInfo s
    * @param matrix The input matrix.
    * @return The output matrix.
    */
-  private static BitMatrix convertByteMatrixToBitMatrix(ByteMatrix matrix) {
-    int matrixWidgth = matrix.getWidth();
+  private static BitMatrix convertByteMatrixToBitMatrix(ByteMatrix matrix, int reqWidth, int reqHeight) {
```
srowen: It's worth a little javadoc to clarify what these two args are, especially as they're abbreviated

core/src/main/java/com/google/zxing/datamatrix/DataMatrixWriter.java
```diff
@@ -159,17 +160,33 @@ private static BitMatrix encodeLowLevel(DefaultPlacement placement, SymbolInfo s
...
+    int multiple = Math.min(outputWidth / matrixWidth, outputHeight / matrixHeight);
+
+    int leftPadding = (outputWidth - (matrixWidth * multiple)) / 2 ;
+    int topPadding = (outputHeight - (matrixHeight * multiple)) / 2 ;
+
+    BitMatrix output;
+
+    // remove padding if requested width and height are too small
+    if (reqHeight < matrixHeight || reqWidth < matrixWidth) {
+      leftPadding = topPadding = 0 ;
```
srowen: separate these into two statements and remove the space before ;

kent92:
@srowen added the changes you've suggested. Let me know if there's more. Thanks

PR #952, opened 2018-01-27
 Added hint not accompanying ECI for feature phone

kite999:
Since the feature phone does not support ECI with the QR code reading function, it can not read the QR code with ECI.
I added a hint phrase so that I can not give ECI even with ShiftJIS.

### DIFF
--- a/core/src/main/java/com/google/zxing/EncodeHintType.java
+++ b/core/src/main/java/com/google/zxing/EncodeHintType.java
@@ -107,4 +107,10 @@ public enum EncodeHintType {
    * {@link String } value).
    */
   GS1_FORMAT,
+
+  /**
+   * Set whether to add ECI segment of QR code. (type {@link Boolean}, or "true" or "false"
+   * {@link String } value).
+   */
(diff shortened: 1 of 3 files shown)


srowen:
Hm, the problem is the encoded QR code would be wrong. Without the ECI segment, its interpretation will be wrong according to the ISO-8859-1 character set. You may have a niche use case where you need to generate these wrong codes anyway, but I'd suggest you just modify the project for your needs. I don't know if this is general enough to warrant adding to the code.

kite999:
It seems that you can not read the QR code with feature phone in case of shiftjis and bytemode.
It was because ECI exists as a result of verification.
I thought whether I could add a function that does not give ECI.
If you are not using shiftjis and bytemode QR code on feature phone, I will give up.
Certainly it is unlikely that it will be used outside of Japan.

srowen:
If byte mode is being used with Shift_JIS or UTF-8 encoding, it requires an ECI segment. Readers need to read it, or at least, ignore it and make their assumption about the encoding.

PR #956, opened 2018-02-12
Using 'zxing' prefix for all generated artifacts

lgoldstein:
The idea is to have a standard prefix for all _zxing_ artifacts so they don't conflict with existing ones (IMO) `core`, `javase`, `android` are too generic and might "collide" with others. Furthermore, the prefix makes it perfectly clear at a glance what is the function of the deployed JAR/WAR/artifact.

I am aware that this breaks backward compatibility in the sense that _Maven_ POM(s) using them will need to be updated, but in the long run I believe this is the correct way to go - see [Spring](https://spring.io/projects), [Hibernate](http://hibernate.org/) and others that use this convention.

P.S. I also removed the `<version>` where artifact version is same as the parent as it is the default _Maven_ behavior and also makes the POM(s) more robust since we avoid version duplication (a.k.a. the [D.R.Y. principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)). Note that I have left the special version override `4.7.7` in the _android_ POM.

### DIFF
--- a/android-core/pom.xml
+++ b/android-core/pom.xml
@@ -16,9 +16,7 @@
  -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
-
-  <artifactId>android-core</artifactId>
-  <version>3.3.3-SNAPSHOT</version>
+  <artifactId>zxing-android-core</artifactId>
   <packaging>jar</packaging>
 
   <dependencies>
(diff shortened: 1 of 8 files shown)


lgoldstein:
>> It's a non starter for the reason you give.

I understand the reluctance to break backward compatibility, but it has been done in the past - as long as it is for good reasons, AFAIK most professional programmers welcome such a change. Furthermore, it is something very rare thus relatively acceptable.

>> It is already prefixed by the group ID. There is no such collision. I guess you mean using the JAR files individually but you can rename them however you like.

Think about the JAR(s) being included in a WAR - the _maven-war-plugin_ knows how to pack the dependencies in a WAR as-is - renaming them is possible, but adds (IMO) un-necessary complexity to the build process as well as introducing confusion since the JAR name is not WYSIWYG when listed in the WAR after renaming.

In view of this I would like ask to re-consider this PR...

srowen:
I just don't see that as a good reason for a pretty non-trivial build inconvenience. Other plugins lose the ability to understand these are the same artifact; you can now accidentally include two copies of the same library because the dependency resolver doesn't know they're the same. 

There's not even a practical example of the theoretical problem here. As you say, people and plugins know how to deal with this situation anyway.

PR #958, opened 2018-02-15
Fix #957 - Provide OSGi Metadata in Manifest

### LINKED ISSUE #957: Provide OSGi Metadata in Manifest
To make it possible to use the lib without any extra work inside OSGi it would be good to include OSGi Metadata in the Manifest. Since this is ignored for no OSGi environments it wont disturb other users.

I'll provide a PR for this.

laeubi:
Add the necessary config to generate OSGi-Metainfos

### DIFF
--- a/core/bnd.bnd
+++ b/core/bnd.bnd
@@ -1,0 +1,20 @@
+# Copyright 2018 ZXing authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Bundle-SymbolicName: com.google.zxing.core
+Import-Package: !com.google.zxing*
+
+-exportcontents: \
+	com.google.zxing*
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -67,24 +67,24 @@
         <artifactId>bnd-maven-plugin</artifactId>
         <version>3.5.0</version>
 
-		<executions>
-			<execution>
-				<goals>
-					<goal>bnd-process</goal>
-				</goals>
-				</execution>
-		</executions>
-		</plugin>
+        <executions>
+        	<execution>
+                <goals>
+                	<goal>bnd-process</goal>
+                </goals>
+                </execution>
+        </executions>
+        </plugin>
 
-		<plugin>
-			<groupId>org.apache.maven.plugins</groupId>
-			<artifactId>maven-jar-plugin</artifactId>
-			<configuration>
-				<archive>
-					<manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
-				</archive>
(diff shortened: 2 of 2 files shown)


srowen:
I'm neutral on it but if it is just a file in META-INF it's not a big deal. Is it really just a static file? Then just the jar plugin config is enough. What does the other plugin do?

laeubi:
The BND plugin generates just some Headers for the MANIFEST.MF file, the bnd.bnd file is not included, just carry the config.
The config for the maven-jar-plugin simply instructs it to take this generated header into acount and merge with the default one.

srowen:
I see. Do you know what the headers are? there are more direct ways with built-in plugins to add headers to the manifest, or add files to META-INF.

laeubi:
Of course you can add the header manually, the advantage of BND is that it commutes required values automatically.

laeubi:
For example Version and exported packages might change then it will be updated automatically.

core/pom.xml
```diff
@@ -60,6 +60,33 @@
...
+				</execution>
+		</executions>
+		</plugin>
+
+		<plugin>
+			<groupId>org.apache.maven.plugins</groupId>
+			<artifactId>maven-jar-plugin</artifactId>
+			<configuration>
+				<archive>
+					<manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
```
srowen: This should be the default, no?

core/pom.xml
```diff
@@ -60,6 +60,33 @@
...
     </profile>
   </profiles>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>biz.aQute.bnd</groupId>
+        <artifactId>bnd-maven-plugin</artifactId>
+        <version>3.5.0</version>
+
+		<executions>
```
srowen: This should use 2-space indent consistently.

core/bnd.bnd
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Bundle-SymbolicName: com.google.zxing.core
+Import-Package: !com.google.zxing*
```
srowen: Does it matter that core isn't the only artifact that exports packages in this namespace? (For example this is why the Java 9 module system won't work here).

core/pom.xml
laeubi: In my tests the maven jar plugin ignores it otherwise :-(

core/bnd.bnd
laeubi: Not for this use case.

core/pom.xml
laeubi: I replaced tabs by spaces

core/bnd.bnd
laeubi: Just wondering does other modules implement classes in the same packages as the core bundle?

srowen: Yes, from years and years ago. `javase` has some classes in this package too.

core/pom.xml
```diff
@@ -67,24 +67,24 @@
...
-		<executions>
-			<execution>
-				<goals>
-					<goal>bnd-process</goal>
-				</goals>
-				</execution>
-		</executions>
-		</plugin>
+        <executions>
+        	<execution>
```
srowen: This is still using tabs I think, and indented too deeply.

PR #970, opened 2018-03-08
Make integration of the core to 3rd party Android apps simpler 

serso:
Zxing has an Android app implemented in `android` submodule. It contains classes that might be used by other developers who want to implement a QR/bar code scanner in their apps. The simplest way to do it now is to copy sources of this sumbodule and modify them if needed. This is not very good as:

1. The submodule is quite big and contains a lot of sources (e.g. `PreferencesActivity`) that are unlikely to be needed in other apps

2. The submodule contains some code that is better to be shared as it is quite generic (e.g. `ViewfinderView`, `DecodeHintManager`, CameraConfigurationManager`) and might be updated in the future to workaround device specific bugs

To avoid this developers should be able to import only generic Android classes in their projects.

This pull request divides `android` classes into two groups: zxing-specific classes and generic classes. The latter are moved into `android-core` submodule and can be used by other developers via a simple dependency import.


srowen:
It's an app, as you say, and not intended to be reused directly. I don't think the UI elements should be presented as things to reuse directly, mostly because this has been widely abused over time.

The parts that are reusable are in `core` and `android-core`. I think it's possible that some more logic could move there for the benefit of other apps, sure. I don't want to make any breaking changes at this point though.

As such I don't think it can change to an `aar` module (but if there's a way to publish as `jar` as well that's fine) and would only refactor logic, not UI elements. For example, logic about figuring out preview size is fine, but not colors.

serso:
thank you for a prompt response, @srowen. Is packaging a bigger issue for you than having e.g. `ViewfinderView` in `android-core`? I think it should be possible to package `android-core` into a jar by moving all resources back to `android`.

Rephrasing: what classes you definitely don't want to see in `android-core`? Or is it okay to have all classes that were moved given that `android-core` is packaged into a jar?

srowen:
I think the Android resources should probably stay in `android`, yes. Views, manifests, resources, all that. I doubt they're very reusable. Some logic in utility classes could move like picking a preview size, or configuring the camera, maybe.

serso:
ok, so it looks like this pull-request is a no go. A goal was to make it simple for other devs to integrate ZXing in their apps with minimum code which implies providing some UI classes (as `ViewfinderView`, `DecodeThread`, etc).

I'll try to maintain my fork though. If there are people interested in it, they can build it and use it as a local dependency.

serso:
@srowen, another proposal, maybe it's a good compromise?

Instead of moving UI classes to `android-core`, we can move them to a dedicated submodule `android-ui`. That would mean no changes to `android-core` and minimal changes to `android` (like ones in commits 1-2 above). After this, submodule `android` would depend on `android-ui` and maybe on `android-core` and submodule `android-ui` - on `android-core`.

Those, who would want to reuse UI classes, can import `android-ui` in their projects. These classes would be self-sufficient (and unlikely to break) as they wouldn't depend on activities from `android`.

Do you think it's a good idea?

srowen:
That makes more sense, but the issue here is that I actually don't want the classes to be reused as a library. Part of it is that it simply wasn't designed well for that purpose, and part is that it has been very widely abused already. People just clone the exact look-feel of the app and people think they're using this app when they're using something else. I'd rather people develop their own interfaces for this reason.

serso:
Not sure I agree with you. People who want to abuse the library will abuse it by copying the sources and releasing clone apps. But developers who want to integrate it to their projects as an additional (small) feature have to reinvent a bicycle and write a boilerplate code (`android-integration` is not an option for everyone).

I'm closing the pull request. If you change your mind feel free to reopen it :)

srowen:
That is true, but packaging it for reuse implies it is encouraged, and it isn't. It's not just that users think they're using this app when they're not, but, the UI design is very old anyway (pre-dates Android 1.0). As I say, not against pulling out utility code, as that is still 'relevant' and does not present the same user confusion issue.

serso:
> the UI design is very old anyway (pre-dates Android 1.0)
`ViewfinderView` implements a very simple and solid UI which can be reused as-is. I didn't copy `CaptureActivity` which additionally contains a result view and a toolbar.

PR #973, opened 2018-03-14
Improve support for Macro PDF417

gitu:
I had troubles reading a PDF417 barcode with multiple symbols. These PR would fix that. It is though not reverse compatible regarding the optionalData field on the ResultMetadata.


core/src/main/java/com/google/zxing/pdf417/PDF417ResultMetadata.java
```diff
@@ -42,14 +56,26 @@ public void setFileId(String fileId) {
...
+  @Deprecated
   public int[] getOptionalData() {
-    return optionalData;
+    return null;
   }
 
+  /**
+   * @deprecated no replacement
+   */
+  @Deprecated
```
srowen: Is it not possible to continue to support this at all? just leave it also populated? I'd hate to introduce any breaking changes.

core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
```diff
@@ -172,35 +179,58 @@ private static int decodeMacroBlock(int[] codewords, int codeIndex, PDF417Result
     codeIndex = textCompaction(codewords, codeIndex, fileId);
     resultMetadata.setFileId(fileId.toString());
 
-    switch (codewords[codeIndex]) {
-      case BEGIN_MACRO_PDF417_OPTIONAL_FIELD:
-        codeIndex++;
-        int[] additionalOptionCodeWords = new int[codewords[0] - codeIndex];
-        int additionalOptionCodeWordsIndex = 0;
+    while (codeIndex < codewords[0]) {
+        switch (codewords[codeIndex]) {
```
srowen: The project uses 2-space indent

core/src/main/java/com/google/zxing/pdf417/PDF417ResultMetadata.java
gitu: We could just populate the array after the parsing. But there are also no tests for that functionality. I'll give it a try.

srowen: Yeah I'd leave the functionality. Deprecating it is OK. This kind of has a replacement, right? the new fields you're adding have some of the same info?

core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
gitu: I added also an .editorconfig, that way i didn't had to fix the identation manually.

core/src/main/java/com/google/zxing/pdf417/PDF417ResultMetadata.java
gitu: Exactly they now consume the array, the behavior now is not exactly the same as it parses the fields and allows for multiple ones to be set. For symbols that were parsed before the data found in the array should be the same.

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
```
srowen: Oh sure, what reads this? Eclipse et al? that's fine, those settings look standard and like what the project uses.

gitu: eclipse needs a plugin, but some others support it by default: http://editorconfig.org/

core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
```diff
@@ -155,8 +164,8 @@ static DecoderResult decode(int[] codewords, String ecLevel) throws FormatExcept
     return decoderResult;
   }
 
-  private static int decodeMacroBlock(int[] codewords, int codeIndex, PDF417ResultMetadata resultMetadata)
-      throws FormatException {
+  protected static int decodeMacroBlock(int[] codewords, int codeIndex, PDF417ResultMetadata resultMetadata)
```
srowen: Last nit: if this is just visible for testing, it can probably be package-private only? 
I had also done that extra 4-space indent for wrapping continuations like the 'throws' clause on purpose, but, I don't really care.

gitu: sure no problem; will change the visibility

I'm lazy so I just use the auto layout from IDEA, which works usually quite well, but doesn't indent like that, there are additional changes like that, shall I change them back as well?

srowen: If you're editing this line anyway, maybe change it back, but wouldn't worry about undoing every whitespace change.

gitu: I changed it back mostly, makes the change a bit cleaner, shall i rebase or will you do a squash merge?

PR #1020, opened 2018-06-02
String of zero polynomial be "0"

makiuchi-d:
GenericGFPoly.toString() returns empty string (`""`) when that is zero.
It should be `"0"`.


srowen:
Any chance you can add a test that covers this? Should just be one line or so.

srowen:
Hm! I didn't realize I hadn't even written a single test for this class already. This is fine. I'll probably then move it to another standalone test and add a few for good measure.

Issue #996, opened 2018-04-20
Oreo icon shape


c-e-p-x-u-o:
Please add support for icon shape
https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive.html
https://material.io/guidelines/style/icons.html

srowen:
If you're able to design and submit a new icon scheme in a PR, I'll take a look, but otherwise not going to happen.

c-e-p-x-u-o:
Really it's not a design but just a shape of the icon's background.

c-e-p-x-u-o:
But I'll try to help, as you ask.

srowen:
Yeah, though I figure just chopping a circle from the icon wouldn't look great. Maybe surround it with a circle but then it gets small. It's for better or worse such a well known icon that I'm reluctant to change it, but, if it's a pretty good and recognizable take on the existing icon, could be fine.

c-e-p-x-u-o:
ok, I'll try to make a concept

marescamichael2:
Thanks

PR #1026, opened 2018-06-13
Support Adaptive Icons & Fixed strings.xml Korean

kdwoo1217:
#996 I think he want to like this called Adaptive Icons on Android Oreo.
And I Fixed strings.xml Korean because some sentences are too stiff to Korean.
I don't know this is right PR.
I read your Release Process in here [https://github.com/zxing/zxing/wiki/Getting-Started-Developing](url).
But I don't know what it means....
If this PR is wrong the send me E-Mail or comment here. Thank you.

screenshot_20180614-005911
screenshot_20180614-005928
screenshot_20180614-010001


srowen:
This has a number of superfluous XML changes that need to be reverted. Also please separate out the Korean strings change. Finally, do you really need new copies of the icon? can the adaptive icon not just mask the existing icon somehow? maybe not

kdwoo1217:
I just fixed AndroidManifest.xml in zxing/anroid/AndroidManifest.xml and Add ic_launcher files in zxing/android/res/mipmap folders.
Adaptive Icon is If I want change to possible icon background using launcher settings like that.
That is separate icon foreground and background.

kdwoo1217:
And how to separate Korean strings change? 
I do not know because I am a student who has just used Git.

srowen:
I think you'll want to start over and delete this pull request and branch. Make a branch with just the Korean strings change, push it, and open a PR for that branch. Make another branch with the icon change only, push it, make a PR for that branch. Before pushing, ensure you do not commit any other changes to other files. You can see here than about 70 XML files had their "UTF-8" declaration changed.

PR #1028, opened 2018-06-13
Adaptive Icon Support

kdwoo1217:
#996 I think he want to like this called Adaptive Icons on Android Oreo.
I don't know this is right PR.
I read your Release Process in here [https://github.com/zxing/zxing/wiki/Getting-Started-Developing](url).
But I don't know what it means....
If this PR is wrong the send me E-Mail or comment here. Thank you.

screenshot_20180614-005911
screenshot_20180614-005928
screenshot_20180614-010001


android/AndroidManifest.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
srowen: This part of the change should still be reverted.

android/res/values-fa/strings.xml
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
```
srowen: Likewise this

android/res/xml/preferences.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
```
srowen: And this

kdwoo1217:
Adaptive icon is show like this on Android 8.0.
adaptive-icons

PR #1030, opened 2018-06-13
Create Barcode Scanner Build Docs Korean

kdwoo1217:
I try to build this android application. This is so hard to me.
I think another Korean Student feel like me.
So I make How to Build this application using MD.
Check Please


srowen:
Thank you for this. However I think it isn't possible to maintain these documents in Korean too, as I can't maintain them myself. The build docs are currently in the wiki anyway, not in docs/. However you can maintain this documentation on any personal blog or Google Docs.

PR #1034, opened 2018-06-19
Add basic support for JSR 376 modules

wienczny:
Add Automatic-Module-Name to core and javase
Move Translator files to their own module as a single package can only be contained within one module.

### DIFF
--- a/javase/pom.xml
+++ b/javase/pom.xml
@@ -16,7 +16,7 @@
  -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  
+
   <artifactId>javase</artifactId>
   <version>3.3.4-SNAPSHOT</version>
   <packaging>jar</packaging>
@@ -64,6 +64,17 @@
           <descriptorRefs>
             <descriptorRef>jar-with-dependencies</descriptorRef>
           </descriptorRefs>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>com.google.zxing.client.j2se</Automatic-Module-Name>
+            </manifestEntries>
+          </archive>
         </configuration>
       </plugin>
     </plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -73,6 +73,7 @@
   <modules>
     <module>core</module>
     <module>javase</module>
+    <module>translator</module>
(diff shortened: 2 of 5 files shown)


srowen:
You've hit on the problem here - different modules define classes in the same package and that won't work with the module mechanism. Yes you'd have to move classes around. I'm reluctant to do it as it's a breaking change. Some of these classes aren't really APIs, I think a few are? At least that's why I didn't see enough value in doing this. Don't even think it needs a new Maven module. Maybe one day will release a 4.x release that makes a bunch of little breaking changes including all this.

wienczny:
I tried to use the Java SE module with Java 9+ which does not work. The automatic module naming and duplicate packages are are problem here. "core" is not a good Java module name ;-) I did not touch the Android parts as I don't expect to see JSR-376 on Android any time soon.
For JavaSE there are just two classes that don't fit into the picture: HtmlAssetTranslator and StringsResourceTranslator. Because they are in com.google.zxing they break JavaSE 9+ comptibility.
For me those classes look like they are used to build the Android translations. Because there are no dependencies on any other zxing class, I moved them to their own Maven module.
My patch only adds basic support for JSR-376 using a backwards compatible fallback. If you wanted full support you would have to add a module-info.java to the source root of any maven module. This has two drawbacks: You need Java 9+ for compilation of this file and the resulting .class has a new byte code version which breaks old versions of asm.

srowen:
core is a fine Maven artifact ID, but, this was all conceived in 2007 and would definitely have been organized differently today. 

Yes I understand the problem and fix, but am saying it causes some minor potential breakage. I think there is a little more breakage needed to make all the modules compliant (excluding Android).

But these modules work fine in Java 9 and 10, they just don't take advantage of checks on what classes modules export. But we'd be exporting most every class anyway for backwards compatibility. And yes the idea is to maintain Java 7 bytecode compatibility. 

So I don't think this really adds much, which is why I hadn't done it already. For a major version update? Could see overhauling all the modules, sure, but, even that may be unuseful given that this project is already long in maintenance mode.

wienczny:
My problem is the modules only work in Java9+ projects that don't use modules or include core only.

> [INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] error: the unnamed module reads package com.google.zxing from both core and javase
[ERROR] error: module javase reads package com.google.zxing from both core and javase
[ERROR] error: module slf4j.api reads package com.google.zxing from both core and javase
[ERROR] error: module core reads package com.google.zxing from both javase and core
[ERROR] error: module guava reads package com.google.zxing from both core and javase

srowen:
It works fine in Java 9 without any of these changes. Without these declarations they are assumed to export all classes. What are you doing with the stock build that causes this?

wienczny:
I've prepared a minimal demo at https://github.com/wienczny/zxing-demo
The problem only occurs when the importing Maven project is a JSR-376 module. 
Removing the module-info.java makes the project compile again.

srowen:
Yes, don't add module-info.java and it's fine. No change is required at all to work in Java 9. Right?

srowen:
Ok yeah for reasons above I would leave this for now. It doesn't make anything work better.

PR #1039, opened 2018-06-30
Check if characters are digits for EAN_13, EAN_8, ITF and UPC_EAN_EXTENSION

keoni29:
Several barcode formats only support digits 0-9, namely

- BarcodeFormat.EAN_13 
- BarcodeFormat.EAN_8 
- BarcodeFormat.ITF 
- BarcodeFormat.UPC_EAN_EXTENSION

Passing different characters to the encode function causes a runtime exception. 

**Proposed solution**
Check return value of Character.digit and throw IllegalArgumentException if it is -1 (invalid character.)
Source: https://developer.android.com/reference/java/lang/Character#digit(int,%20int)

### DIFF
--- a/core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
+++ b/core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
@@ -23,6 +23,7 @@ import com.google.zxing.WriterException;
 import com.google.zxing.common.BitMatrix;
 
 import java.util.Map;
+import java.util.regex.Pattern;
 
 /**
  * <p>Encapsulates functionality and implementation that is common to one-dimensional barcodes.</p>
@@ -90,6 +91,7 @@ public abstract class OneDimensionalCodeWriter implements Writer {
     return output;
   }
 
+  private static final Pattern NUMERIC = Pattern.compile("[0-9]+");
 
   /**
    * @param target encode black/white pattern into this array
--- a/core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
+++ b/core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
@@ -93,6 +93,10 @@ public abstract class OneDimensionalCodeWriter implements Writer {
 
   private static final Pattern NUMERIC = Pattern.compile("[0-9]+");
 
+  protected static boolean checkNumeric(String contents) {
+    return NUMERIC.matcher(contents).matches();
+  }
+
   /**
    * @param target encode black/white pattern into this array
    * @param pos position to start encoding at in {@code target}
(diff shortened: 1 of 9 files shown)


srowen:
Rather than check every call to `Character.digit`, how about checking the input once at the start of the various `encode` methods with a regex? You can make a utility method in `UPCEANWriter` maybe.

```
private static final Pattern NUMERIC = Pattern.compile("[0-9]+");
...
if (!NUMERIC.matcher(contents).matches()) {
  ...
} 
```

Lastly if you have a moment to add a simple test of these behaviors that would really complete it.

core/src/test/java/com/google/zxing/oned/EAN8WriterTestCase.java
```diff
@@ -42,4 +42,14 @@ public void testAddChecksumAndEncode() throws WriterException {
     assertEquals(testStr, BitMatrixTestCase.matrixToString(result));
   }
 
+  @Test
+  public void testEncodeIllegalCharacters() throws WriterException {
```
srowen: You should be able to write `@Test(expected = WriterException.class)` and then the body of the test is just the line that causes the exception.

core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
```diff
@@ -90,6 +91,7 @@ private static BitMatrix renderResult(boolean[] code, int width, int height, int
     return output;
   }
 
+  private static final Pattern NUMERIC = Pattern.compile("[0-9]+");
```
srowen: Can you write a simple `checkNumeric` method here or something and then just call it in each of the subclasses? that way the invocation isn't duplicated.

srowen:
@keoni29 if you'll make the last updates, I'll merge

keoni29:
Have not had time this week to look at it. I can probably get it done this weekend. Thanks for your suggestions.

core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
```diff
@@ -90,6 +92,9 @@ private static BitMatrix renderResult(boolean[] code, int width, int height, int
     return output;
   }
 
+  protected static boolean checkNumeric(String contents) {
+    return NUMERIC.matcher(contents).matches();
```
srowen: Why not just have this method check whether it matches and throw the exception? that way you don't have to repeat the check and throw everywhere. It can be `final` too.

core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
```diff
@@ -90,6 +92,14 @@ private static BitMatrix renderResult(boolean[] code, int width, int height, int
     return output;
   }
 
+  /**
+   * Throw IllegalArgumentException if input contains characters other than digits 0-9.
+   */
+  protected static final void checkNumeric(String contents) throws IllegalArgumentException{
+    if (!NUMERIC.matcher(contents).matches()) {
+        throw new IllegalArgumentException("Input should only contain digits 0-9");
```
srowen: The indentation is off here and in a test below but maybe I can just adjust it later. Change looks OK.

keoni29: Resolved with latest commit

PR #1044, opened 2018-07-21
fix binary shift cost heuristics

aniederle:
This introduces some tests that show that the current heuristic that determines whether an encoding state can savely been dropped isn't valid for binary shift mode. That flaw may lead to inefficent encoding for non-textual data, i.e. the encoding uses some more bytes than the original octet encoding does.

It also comes with the fix for that heuristic.

### DIFF
--- a/core/src/main/java/com/google/zxing/aztec/encoder/State.java
+++ b/core/src/main/java/com/google/zxing/aztec/encoder/State.java
@@ -170,15 +170,16 @@ final class State {
   }
   
   private static int calculateBinaryShiftCost(State state) {
-      if (state.binaryShiftByteCount > 62) {
-          return 21; // B/S with extended length
-      } else if (state.binaryShiftByteCount > 31) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/zxing/aztec/encoder/State.java
```diff
@@ -165,5 +168,17 @@ BitArray toBitArray(byte[] text) {
   public String toString() {
     return String.format("%s bits=%d bytes=%d", HighLevelEncoder.MODE_NAMES[mode], bitCount, binaryShiftByteCount);
   }
+  
+  private static int calculateBinaryShiftCost(State state) {
+      if (state.binaryShiftByteCount > 62) {
```
srowen: Nit: use 2-space indent here. Also I think you can dispense with the "else" statements and just write:

```
if (...) {
   return ...;
}
if (...) {
   return ...;
}
```

PR #1046, opened 2018-07-29
Directory Bug

FlXME:
The correct pattern to make a directory is: `if (!dir.mkdirs() && !dir.isDirectory()) { error }` mkdirs checks for exists so the exists check is redundant.


srowen:
OK by me. Looks like `mkdirs()` doesn't throw an error if it already exists, so guarding with `exists()` isn't important. Adding a double-check of the dir creation seems OK.

PR #1047, opened 2018-07-29
Directory Bug

FlXME:
The correct pattern to make a directory is: `if (!dir.mkdirs() && !dir.isDirectory()) { error }` mkdirs checks for exists so the exists check is redundant.


srowen:
Closing as a duplicate of #1046

Issue #1090, opened 2018-10-19
Getting an incorrect image from the camera Meizu MX4 Pro


m4xp1:
Barcode scanner does not work on Meizu MX4 Pro. This is related to the following lines in the CameraConfigurationManager:

```Java
//SetRecordingHint to true also a workaround for low framerate on Nexus 4
//https://stackoverflow.com/questions/14131900/extreme-camera-lag-on-nexus-4
parameters.setRecordingHint(true);
```

For some unclear reason, this setting forces the cameras to provide broken images as in the screenshot below (bottom right corner).

<img width="308" alt="screen shot 2018-10-19 at 19 54 54" src="https://user-images.githubusercontent.com/27896818/47234033-24986380-d3dd-11e8-94aa-fbf736df0c51.png"><img width="308" alt="screen shot 2018-10-19 at 19 54 54" src="https://user-images.githubusercontent.com/27896818/47234012-0f233980-d3dd-11e8-936d-18d0e102cf85.png">

I created a pull request to delete these lines. Although it is possible to add the correct checks and perform them only for Nexus 4. But I have a weekly device to test the verification code and I don’t know if the remote code affects any other devices. I would like to get more information.

Finally! Thanks for the wonderful library and application, I constantly use it, it is the best!

PR #1091, opened 2018-10-19
Getting an incorrect image from the camera Meizu MX4 Pro #1090

m4xp1:


srowen:
No, this was added specifically to improve decoding. If a device has a problem with it it's a device bug.

PR #1098, opened 2018-10-27
 Avoid ArrayIndexOutOfBoundsException in BitMatrix.get()

makiuchi-d:
The pixel on out side of the image can be assumed white.

ref: [makiuchi-d/gozxing/pull/9](/makiuchi-d/gozxing/pull/9)


srowen:
The point here is to not check bounds here for speed. This is supposed to be avoided more efficiently by callers.

PR #1101, opened 2018-11-01
fix code93 extended character decoder

makiuchi-d:
Some characters (`(%)U`, `(%)V` and `(%)W` map to `NUL`, `'@'` and ``'`'``) were decoded to wrong value `DEL` in Code93Reader.


srowen:
That doesn't seem to agree with https://en.wikipedia.org/wiki/Code_93#Full_ASCII_Code_93 ? I think that's all I have been looking at. What ref are you looking at?

makiuchi-d:
I referenced same wikipedia's page too.
That table clearly shows:
* NUL mapped to (%)U
* '@' mapped to (%)V
* '`' mapped to (%)W
* DEL mapped to only (%)T, (%)X, (%)Y and (%)Z

srowen:
Yeah you're right, I think I misread this again for a second time. I agree with the change, thanks!

PR #1103, opened 2018-11-04
Optimize code width calculation in Code39Writer

makiuchi-d:
Every code39 character has 9 bars containing 3 wide bars, and separated by a white bar.
`(9-3)*1 + 3*2 + 1 = 13`


srowen:
Yes, looks good.

PR #1104, opened 2018-11-07
Support Full-ASCII in Code93Writer

makiuchi-d:
Support full ASCII encoding in Code93Writer and remove redundancy.


makiuchi-d:
In Code39, for example, "/" has two ways to encode, "/" in normal mode and "/O" in extended mode.
This is a reason to need to specify whether extend-mode or not by the caller, as you said.

In Code93, different from Code39, control characters are mapped to specific symbols.
ex. "/" = 40, (/) = 45.
There is no confusion like Code39.

I add a test to check that the non-extendable characters (0-9A-Z-. $/+%) are not modified by the convertToExtended method.

core/src/main/java/com/google/zxing/oned/Code93Writer.java
```diff
@@ -40,74 +40,49 @@ public BitMatrix encode(String contents,
...
-  /**
-   * @param target output to append to
-   * @param pos start position
-   * @param pattern pattern to append
-   * @param startColor unused
-   * @return 9
-   * @deprecated without replacement; intended as an internal-only method
-   */
-  @Deprecated
-  protected static int appendPattern(boolean[] target, int pos, int[] pattern, boolean startColor) {
```
srowen: This shouldn't have been protected in the first place, or else yes I'd have removed it. I kept it to maintain full backwards-compatibility, even if I don't see any reason for a caller to use this. There is a check in the build that I thought would fail if a protected or public method was removed, but, this passed. I still think maybe it's best to keep the method, just to minimize change and to remove any possibility of a maintenance release breaking some code.

core/src/main/java/com/google/zxing/oned/Code93Writer.java
```diff
@@ -40,74 +40,49 @@ public BitMatrix encode(String contents,
 
   @Override
   public boolean[] encode(String contents) {
+    contents = convertToExtended(contents);
```
srowen: We probably need a comment in the javadoc here explaining that the contents should _not_ already be encoded for extended characters. Otherwise this would double-encode them.

core/src/main/java/com/google/zxing/oned/Code93Writer.java
```diff
@@ -40,74 +40,49 @@ public BitMatrix encode(String contents,
...
     int length = contents.length();
     if (length > 80) {
       throw new IllegalArgumentException(
-        "Requested contents should be less than 80 digits long, but got " + length);
+        "Requested contents should be less than 80 digits long after extended, but got " + length);
```
srowen: "... after converting to extended encoding, ..." ?

core/src/main/java/com/google/zxing/oned/Code93Writer.java
```diff
@@ -125,4 +100,66 @@ private static int computeChecksumIndex(String contents, int maxWeight) {
...
+    int length = contents.length();
+    StringBuilder extendedContent = new StringBuilder(length * 2);
+    for (int i = 0; i < length; i++) {
+      char character = contents.charAt(i);
+      if (character == 0) {
+        // NUL: (%)U
+        extendedContent.append("bU");
+      } else if (character <= 26) {
+        // SOH - SUB: ($)A - ($)Z
+        extendedContent.append('a');
```
srowen: I thought these were prefixed with $? or is https://en.wikipedia.org/wiki/Code_93#Full_ASCII_Code_93 not the right reference?

makiuchi-d: Please see `Code93Reader.ALPHABET_STRING`.
Control chars are mapped to `abcd`.

I add a comment in this method.

core/src/test/java/com/google/zxing/oned/Code93WriterTestCase.java
```diff
@@ -36,11 +37,38 @@ public void testEncode() throws WriterException {
...
   }
 
   private static void doTest(String input, CharSequence expected) throws WriterException {
     BitMatrix result = new Code93Writer().encode(input, BarcodeFormat.CODE_93, 0, 0);
     assertEquals(expected, BitMatrixTestCase.matrixToString(result));
   }
 
+  @Test
+  public void testConvertToExtended() throws Exception {
+    Method method = Code93Writer.class.getDeclaredMethod("convertToExtended", String.class);
```
srowen: Oh, you can make it package-private for testing rather than use reflection

core/src/main/java/com/google/zxing/oned/Code93Writer.java
```diff
@@ -38,60 +38,51 @@ public BitMatrix encode(String contents,
     return super.encode(contents, format, width, height, hints);
   }
 
+  /**
+   * @param contents barcode contents to encode. it should not be encoded for extended characters.
```
srowen: Nit: "It should not be ..."

PR #1114, opened 2018-12-05
Avoid book search crash on invalid intent

johnwhall:
Reproducible with

`Intent i = new Intent("com.google.zxing.client.android.SEARCH_BOOK_CONTENTS");
startActivity(i);`


srowen:
Yeah that's fine. I don't think I can update the app anymore though.

PR #1130, opened 2019-02-04
Fixing checksum error in UPC-E barcodes

fitchfuloffun:
When generating a UPC-E barcode, the code checks whether the provided value has 7 or 8 digits: 

- If it has 7, it provides and appends the correct check digit, by first converting the UPC-E value to a UPC-A value. 

- If it has 8, it checks to see that the provided check digit matches with what Zxing's code would have assigned it. This will fail if provided the correct check digit, as the code was not testing with a converted UPC-A example. 

This issue has been fixed with a small change.


srowen:
Looks good, yeah. I double-checked EAN-8 and looks like it doesn't need a similar change.

PR #1146, opened 2019-03-11
Add zh-rTW translations.

s8321414:
I added zh-rTW translations, please review and merge it :smiley:

### DIFF
--- a/android/res/values-zh-rTW/strings.xml
+++ b/android/res/values-zh-rTW/strings.xml
@@ -1,0 +1,143 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright (C) 2014 ZXing authors
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


srowen:
Pardon my ignorance, but is this a simple or traditional Chinese translation? is it not already covered by zh or zh-rCN?

I can't update the Android app any further, so there's no point in updating the app from here anyway.

s8321414:
@srowen This is Traditional Chinese (Taiwan) and not covered by zh and zh-rCN, in fact, zh can't show on a zh-rTW Android system...

srowen:
Hm, OK, so there are supposed to be 3 variants? I can merge it but it can't be released.

srowen:
If you are reacting to 'can't be released' -- the app can't be published to the app store anymore, so I don't know if it matters what we update here.

I thought `zh` was already a traditional Chinese translation, but of course I have no knowledge of the language myself. Why wouldn't a `zh-rTW` device match it? if the translation is the same, we can symlink if needed.

s8321414:
I have no idea why, but I can see some terms in zh are different from zh-rTW.

Yairimedinac4621:
Thank you

Issue #1157, opened 2019-04-10
can't read qrcode from the image


Donaldhan:
[qrcode](https://loan-star-moon-test.nos-jd.163yun.com/1554796580178101002)

the image contain qrcode, but zxing can't find, anyone can help me ??
the code as follwing:

```
@Slf4j
public class QRCodeUtilTest {
    private static final String DEAFULT_CHARSET = "UTF-8";
    private static final int DEAFULT_SIZE = 125;
    private static final String DEAFULT_FILE_TYPE = "png";

    public static void main(String[] args) throws IOException {
            //https://loan-star-moon-test.nos-jd.163yun.com/1554796580178101002 is the image url
            log.info("image Data read from QR Code: {}", readQRCodeFromURL("https://loan-star-moon-test.nos-jd.163yun.com/1554796580178101002"));
    }
    /**
     * @param qrFile
     * @return
     */
    public static String readQRCode(File qrFile) {
        return readQRCode(qrFile, DEAFULT_CHARSET);
    }

    /**
     * @param qrFile
     * @param charset
     * @return
     * @throws IOException
     * @throws NotFoundException
     */
    public static String readQRCode(File qrFile, String charset) {
        Map<EncodeHintType, ErrorCorrectionLevel> hintMap = new HashMap<EncodeHintType, ErrorCorrectionLevel>();
        hintMap.put(EncodeHintType.ERROR_CORRECTION, ErrorCorrectionLevel.L);
        return readQRCode(qrFile, charset, hintMap);
    }

    /**
     * @param qrFile
     * @param charset
     * @param hintMap
     * @return
     * @throws FileNotFoundException
     * @throws IOException
     * @throws NotFoundException
     */
    public static String readQRCode(File qrFile, String charset, Map hintMap) {
        try {
            BinaryBitmap binaryBitmap = new BinaryBitmap(new HybridBinarizer(
                    new BufferedImageLuminanceSource(
                            ImageIO.read(new FileInputStream(qrFile)))));
            Result qrCodeResult = new MultiFormatReader().decode(binaryBitmap, hintMap);
            return qrCodeResult.getText();
        } catch (Exception e) {
            log.error("QRCodeUtil readQRCode error, qrFile:{}", qrFile.getName(), e);
        }
        return null;
    }

    /**
     * @param imageUrl
     * @return
     */
    public static String readQRCodeFromURL(String imageUrl) {
        File qrFile = null;
        try {
[...]

Issue #1144, opened 2019-03-05
Bug in QRcode/Detector/ FinderPattern[] selectBestPatterns()


Quatrus:
### Bug Reports

624      for (int i = 0; i < possibleCenters.size() && possibleCenters.size() > 3; i++) {
625        FinderPattern pattern = possibleCenters.get(i);
626        if (Math.abs(pattern.getEstimatedModuleSize() - average) > limit) {
627          possibleCenters.remove(i);
628          i--;
629        }
630      }
I am pretty sure that line 626 is wrong. As written it removes the GOOD patterns.
So it should be 
 if (Math.abs(pattern.getEstimatedModuleSize() - average) < limit) {

srowen:
No, it's checking whether the module size in the candidate is too different from the average. The check is correct. You can verify by trying your change; lots of tests fail.

Quatrus:
Yes, sorry, I was too fast with the conclusions. The point is that I have a big set of test images and on some of them this test is wrong. I work with the C++ version and modified the code the following way
```  if (startSize > 3) {
    // But we can only afford to do so if we have at least 4 possibilities to choose from
    int   total = 0;      //<-----new
    float totalModuleSize = 0.0f;
    float average,stdDev,square = 0.0f;
    for (std::size_t i = 0; i < startSize; i++) {
      float size = possibleCenters_[i]->getEstimatedModuleSize();
      if (size != 1.0f)
      {
        totalModuleSize += size;
        square += size * size;
        total++;   //<-----new
      }
    }
    if(total)     //<-----new
    {
       average = totalModuleSize / total;   //<-----new
       stdDev = (float)std::sqrt(square / total - average * average); //<-----new
    }
    else
    {
       average = 1; //<-----new
       stdDev  = 0; //<-----new
    }

    sort(possibleCenters_.begin(), possibleCenters_.end(), FurthestFromAverageComparator(average));
    
    float limit = max(0.2f * average, stdDev);

    for (std::size_t i = 0; i < possibleCenters_.size() && possibleCenters_.size() > 3; i++) {
      if (abs(possibleCenters_[i]->getEstimatedModuleSize() - average) > limit) {
        possibleCenters_.erase(possibleCenters_.begin()+i);
        i--;
      }
    }
  }
```
Now it works MUCH better (for me ;) )

What I did was to exclude the possible centers with suspicious  module size of 1 from the  _average_ and from the _limit_

Quatrus:
[...]

PR #1158, opened 2019-04-15
Improve FinderPatternFinder.selectBestPatterns

makiuchi-d:
I rewrote the `FinderPatternFinder.selectBestPatterns()`.
I got a better score on blackbox test 2, and same scores on other blackbox tests.

In addition, this improvement makes some QR-codes readable which reported in #1157 and #1144.

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
+++ b/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
@@ -640,9 +640,9 @@ public class FinderPatternFinder {
 
           // a^2 + b^2 = c^2 (Pythagorean theorem), and a = b (isosceles triangle).
           // Since any right triangle satisfies the formula c^2 - b^2 - a^2 = 0,
-          // we also need to check both sides separately.
-          // The value of |c^2 - 2 * b^2| + |c^2 - 2 * a^2| increases as the triangle
-          // is dissimilar from isosceles right triangle.
+          // we need to check both two equal sides separately.
+          // The value of |c^2 - 2 * b^2| + |c^2 - 2 * a^2| increases as dissimilarity
+          // from isosceles right triangle.
           double d = Math.abs(squares[2] - 2 * squares[1]) + Math.abs(squares[2] - 2 * squares[0]);
           if (d < distortion) {
             distortion = d;
--- a/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
+++ b/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
@@ -621,21 +621,24 @@ public class FinderPatternFinder {
     FinderPattern[] bestPatterns = new FinderPattern[3];
 
     for (int i = 0; i < possibleCenters.size() - 2; i++) {
-      float minModuleSize = possibleCenters.get(i).getEstimatedModuleSize();
+      FinderPattern fpi = possibleCenters.get(i);
+      float minModuleSize = fpi.getEstimatedModuleSize();
 
       for (int j = i + 1; j < possibleCenters.size() - 1; j++) {
-        double squares0 = squaredDistance(possibleCenters.get(i), possibleCenters.get(j));
+        FinderPattern fpj = possibleCenters.get(j);
+        double squares0 = squaredDistance(fpi, fpj);
 
         for (int k = j + 1; k < possibleCenters.size(); k++) {
-          float maxModuleSize = possibleCenters.get(k).getEstimatedModuleSize();
+          FinderPattern fpk = possibleCenters.get(k);
+          float maxModuleSize = fpk.getEstimatedModuleSize();
           if (maxModuleSize > minModuleSize * 1.4f) {
             // module size is not similar
             continue;
           }
 
           squares[0] = squares0;
-          squares[1] = squaredDistance(possibleCenters.get(j), possibleCenters.get(k));
-          squares[2] = squaredDistance(possibleCenters.get(k), possibleCenters.get(i));
+          squares[1] = squaredDistance(fpj, fpk);
+          squares[2] = squaredDistance(fpi, fpk);
           Arrays.sort(squares);
 
           // a^2 + b^2 = c^2 (Pythagorean theorem), and a = b (isosceles triangle).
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
```diff
@@ -604,85 +613,52 @@ private boolean haveMultiplyConfirmedCenters() {
...
+    double[] squares = new double[3];
+    FinderPattern[] bestPatterns = new FinderPattern[3];
 
-      float limit = Math.max(0.2f * (float) average, stdDev);
+    for (int i = 0; i < possibleCenters.size() - 2; i++) {
+      for (int j = i + 1; j < possibleCenters.size() - 1; j++) {
+        for (int k = j + 1; k < possibleCenters.size(); k++) {
+          float minModuleSize = possibleCenters.get(i).getEstimatedModuleSize();
+          float maxModuleSize = possibleCenters.get(k).getEstimatedModuleSize();
+          if (maxModuleSize > minModuleSize * 1.4f) {
```
makiuchi-d: The limit of the module size in original code was average module size +/- 20%.
So I make a new limit +40% of smallest module size.

core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
```diff
@@ -604,85 +613,52 @@ private boolean haveMultiplyConfirmedCenters() {
...
+            // module size is not similar
+            continue;
+          }
 
-      for (int i = 0; i < possibleCenters.size() && possibleCenters.size() > 3; i++) {
-        FinderPattern pattern = possibleCenters.get(i);
-        if (Math.abs(pattern.getEstimatedModuleSize() - average) > limit) {
-          possibleCenters.remove(i);
-          i--;
+          squares[0] = square(possibleCenters.get(i), possibleCenters.get(j));
```
srowen: It seems like this computation (not assignment) could be lifted out of the inner loop too.

core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
```diff
@@ -604,85 +613,52 @@ private boolean haveMultiplyConfirmedCenters() {
...
+          squares[1] = square(possibleCenters.get(j), possibleCenters.get(k));
+          squares[2] = square(possibleCenters.get(k), possibleCenters.get(i));
+          Arrays.sort(squares);
+
+          double d = Math.abs(squares[2] - 2 * squares[1]) + Math.abs(squares[2] - 2 * squares[0]);
```
makiuchi-d: From the Pythagorean theorem.

srowen: Yes, I get that, but that's not explaining any of the core of the logic. I understand this is looking for an isosceles triangle with sides a, b, and b, and measuring how much a^2 differs from b^2 + b^2. square[2] is the square of the hypotenuse if anything is here, and you compare to the square of both other sides. But why not just measure `Math.abs(squares[2] - squares[1] - squares[0]`? This is the kind of thing that would be useful to write in a comment.

makiuchi-d: Right.
I will write a comment.

makiuchi-d: I

makiuchi-d: I wrote a simple explanation.
Please let me know if not enough.

makiuchi-d: ah...
Please wait, I will rewrite this comment.

makiuchi-d: I rewrote it.
How is this?

core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
```diff
@@ -604,85 +614,60 @@ private boolean haveMultiplyConfirmedCenters() {
...
+    Collections.sort(possibleCenters, moduleComparator);
 
-      Collections.sort(possibleCenters, new FurthestFromAverageComparator((float) average));
+    double distortion = Double.MAX_VALUE;
+    double[] squares = new double[3];
+    FinderPattern[] bestPatterns = new FinderPattern[3];
 
-      float limit = Math.max(0.2f * (float) average, stdDev);
+    for (int i = 0; i < possibleCenters.size() - 2; i++) {
+      float minModuleSize = possibleCenters.get(i).getEstimatedModuleSize();
```
srowen: It may not matter much, but was also thinking you can save the reference to `possibleCenters.get(...)` outside the loop to avoid repeatedly accessing the field and indexing into the collection.

core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
srowen: I think I still have the same question -- this is almost the same as just comparing c^2 to (a^2 + b^2). Would that be more direct and easier to understand?

makiuchi-d: c^2 - (a^2 + b^2) is satisfied by any right triangle, but we need the _isosceles_ right triangle.
for example, a=1, b=√99, c=10 makes it 0. a=7, b=7, c=10 makes it 2.
The best pattern is clearly the latter, but this formula selects the former.

On the case of |c^2 - 2 * b^2| + |c^2 - 2 * a^2|,
a=1, b=√99, c=10 makes it 100.
a=7, b=7, c=10 makes it 4.

So I wrote:
>          // Since any right triangle satisfies the formula c^2 - b^2 - a^2 = 0,
>          // we need to check both two equal sides separately.

core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
makiuchi-d: Do you mean to call toArray before the loop?

core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
srowen: Ah got it, that make sense.

core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
srowen: No, I mean things like:
```
FinderPatternFinder fpi = possibleCenters.get(i);
```
outside the j/k loops for example.
It's not a big deal but this part is performance sensitive.

makiuchi-d: I see. I fixed it.

PR #1162, opened 2019-05-01
Move translator classes to c.g.z.c.j2se.translators and add automatic module header

msgilligan:
This eliminates the split package preventing use of this JAR as a JPMS Module and adds an automatic module header.

I think you're going to reject it because it changes the API, but I'm going to submit it anyway just for feedback and discussion purposes.

### DIFF
--- a/javase/pom.xml
+++ b/javase/pom.xml
@@ -54,6 +54,17 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+              <manifestEntries>
+              <Automatic-Module-Name>com.google.zxing.javase</Automatic-Module-Name>
+            </manifestEntries>
+          </archive>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-assembly-plugin</artifactId>
         <configuration>
           <archive>
--- a/javase/src/main/java/com/google/zxing/client/j2se/translators/HtmlAssetTranslator.java
+++ b/javase/src/main/java/com/google/zxing/client/j2se/translators/HtmlAssetTranslator.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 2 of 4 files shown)


javase/pom.xml
```diff
@@ -52,6 +52,17 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+              <manifestEntries>
```
srowen: Nit: unindent this one step

javase/src/main/java/com/google/zxing/client/j2se/translators/HtmlAssetTranslator.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.zxing;
+package com.google.zxing.client.j2se.translators;
```
srowen: Hm, if it's just these two classes... maybe worthwhile. They are meant to be command-line utilities not so much called from other code. (The docs that show usage of them would have to reflect the new package)

pom.xml
```diff
@@ -501,6 +501,7 @@
             </execution>
           </executions>
         </plugin>
+<!-- Disable API compatibility check since we moved two classes in JavaSE
```
srowen: We can put in an exclusion; I'd have to look up how

msgilligan:
OK @srowen, I'll circle back next week and address your review comments and see if we can figure out how to setup the API compatibility check exclusion.

I'm so happy you're willing to consider this. A few weeks ago I was working on a different project (both projects are JavaFX and bitcoinj-based and use JPMS modules) and didn't think I was going to need the `javase` module. For the new project I need it.

srowen:
@msgilligan I figured out how to make the exclusion, and moved the classes already. you can rebase the rest of this change (or I can just make it if you're busy)

srowen:
I added the rest of the change in https://github.com/zxing/zxing/commit/80e72418a4c29810ca7b63661ea42d19b3a75213 -- thank you.

PR #1173, opened 2019-05-31
Checking the format parameter in the CodaBarWriter.

makiuchi-d:
I found that the format parameter is not checked in `CodaBarWriter.encode()`.
Other oned-writers have a code to check it.


srowen:
I should probably refactor this so subclasses can just say what formats they accept and avoid overriding just to check it.

PR #1176, opened 2019-06-13
Minor change to enable manual selection of the QR code mask pattern

anrcode:
Manual selection of the mask pattern greatly improves performance.


srowen:
Sounds OK to me. It's a little faster, but you may end up with a 'worse' QR code.

PR #1179, opened 2019-06-16
Fix handling of the second value of findFinderPattern() 

makiuchi-d:
The second value of `findFinderPattern()` points the outside pixel of the finder pattern (see [RSS14Reader.java:L301-306](/zxing/zxing/blob/6075d5b2f69f285657ff65b0abcc7c7885a44eed/core/src/main/java/com/google/zxing/oned/rss/RSS14Reader.java#L301-L306)).

I’ｍ going to fix two points in this PR.
1. `ResultPointCallback` parameter will be real center point of the finder pattern.
2.  The _inside char_ will be started from the next pixel of the finder pattern.


srowen:
OK, I don't know this part well, but seems reasonable and improves a test.

PR #1180, opened 2019-06-25
Resolved a crash for Lollipop (API 21) versions while Scanning QR code

usmanrana07:
In Lollipop versions it crashes during scan and gives following error logs:

java.lang.NoSuchMethodError: No interface method sort(Ljava/util/Comparator;)V in class Ljava/util/List; or its super classes


usmanrana07:
Won't this change will avoid crash to let it work for on Lollipop. or we can add a version check on .sort command

srowen:
Just use version 3.3.3 for API < 24.

PR #1237, opened 2019-12-01
Should only use PURE_BARCODE if true.

challgren:
Currently the core code when processing a `DataMatrixReader` or `QRCodeReader` barcode if a user was to set a `DecodedHintType.PURE_BARCODE` to `false`. The decode function would still process the decoding as if `DecodedHintType.PURE_BARCODE` was `true`.


srowen:
That's also reasonable behavior, but this has always been triggered by it being set to any object, and that is in the javadoc too. I wouldn't change the behavior now. I would have made it work this way in retrospect simply because it will active if set to Boolean.FALSE too, which is counterintuitive.

challgren:
Yeah I spent like 3 hours tracking down why our project was having horrible read results of datamatrix barcode and it seems the other developer put pure barcode to false. Probably without reading the javadocs.

PR #1251, opened 2020-01-19
Refactor Member Ignoring Method smell in LuminanceSource class

emaiannone:
Hi, I'm Emanuele Iannone, a master student at University of Salerno.
Since my bachelor's thesis I have been working on a **code smell refactoring plugin** called *aDoctor*, which is able to identify and fix energy-related problems in Android apps.
I launched it on your project, finding different instances of code smells. I chose one of them and let the plugin automatically fix it.
In this case I chose Member Ignoring Method, that is present when a non static method does not use at all instance variables and other non static methods. These kind of smell may have a non trivial impact on energy consumption, as shown in this paper: `https://www.sciencedirect.com/science/article/pii/S0950584918301678`.
Besides, this kind of refactoring does not impact on the functionalities of your app, so it is totally safe. Let me know if you are interested in this refactoring proposal.


srowen:
The changes to all the test source files cause the tests to fail. The contents must be exactly what they are. I think you didn't mean to change that.

The main fix isn't valid, because this method is overridden intentionally. It is polymorphic. You'll probably have to check for that in your plugin.

emaiannone:
Thank you for your feedback. The tool is under development but I'll fix this problem about MIM as soon as possible.Besides, the commited changes is my fault when I did the commit, sorry :(

PR #1275, opened 2020-04-11
Create android.yml

N1992:


srowen:
There is no description of what this does.

srowen:
This also appears to break the RAT check when run in Travis CI

PR #1280, opened 2020-04-30
fix compliation error

gavinsyw:
A compilation error occurs under Android > 5.0 at RSSExpandedReader, line 202.


srowen:
The code definitely requires Java 7 in Android, and Java 8 in core. That is valid in Java 7. I think you may have the wrong target API level. In any event, that has long since compiled and run on Android 5.

PR #1281, opened 2020-05-05
License and copyrights

andrewshadura:
Hi,

This pull request does the following:
 * Specifies the license and copyrights prominently in the README file
 * Updates the copyrights in the main POM file
 * Adds [REUSE metadata](https://reuse.software/spec/) for the currently checked in source code

You may also want to review the information in LICENSE, NOTICES and README and decide whether the historical bits still need to be kept.

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -1,0 +1,31 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: ZXing
+Source: https://github.com/zxing/zxing.git
+
+Files: *
+Copyright: 2007-2020, ZXing authors
+License: Apache-2.0
+
+Files:
+ core/src/main/java/com/google/zxing/datamatrix/encoder/*
+ core/src/test/java/com/google/zxing/datamatrix/encoder/*
+Copyright: 2006-2007, Jeremias Märki
+License: Apache-2.0
+
+Files: docs/js/*
+Copyright: 2005-2014, jQuery Foundation, Inc. and other contributors
+License: MIT
+
+Files: docs/css/apache-maven-fluido-*
+Copyright: 2013, Twitter Inc
+License: Apache-2.0
+
+Files:
+ docs/images/application-certificate.png
+ docs/images/drive-harddisk.png
+ docs/images/icon_info_sml.gif
+ docs/images/internet-web-browser.png
+ docs/images/update.gif
+Copyright: 2003, 2004, Jakub Steiner
+License: CC-BY-SA-2.5 or public-domain
+
--- a/pom.xml
+++ b/pom.xml
@@ -474,6 +474,7 @@
           <version>0.13</version>
           <configuration>
             <excludes>
+              <exclude>.reuse/*</exclude>
               <exclude>**/.*</exclude>
               <exclude>**/.settings/**</exclude>
               <exclude>**/*.iml</exclude>


pom.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
- Copyright (C) 2010 ZXing authors
+ Copyright (C) 2007—2020 ZXing authors
```
srowen, line 3: The date doesn't matter to the copyright. I should have not bothered putting years in, actually, but have done it for all the files with year it was created. It's not worth changing.

README.md
```diff
@@ -88,6 +88,25 @@ library implemented in Java, with ports to other languages.
 Post to the [discussion forum](https://groups.google.com/group/zxing) or tag a question with [`zxing`
 on StackOverflow](https://stackoverflow.com/questions/tagged/zxing).
 
+## License
```
srowen, line 91: This is already in the standard LICENSE file in the repo and also shown by Github; I'd rather not add more to this file to duplicate it.

.reuse/dep5
```diff
@@ -0,0 +1,31 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
```
srowen, line 1: What reads this file? Trying to understand the value in maintaining it too.

pom.xml
andrewshadura: I’d say it’s either worth changing or worth removing, since these copyright notices are used by downstreams to keep track of copyrights and licenses.

srowen: How does the date matter?

README.md
andrewshadura: The LICENSE file is a literal text of the Apache license with an appendix (in my opinion irrelevant and unnecessary) about imagio library (which is not part of the source code and hence doesn’t need to be mentioned).

However, the usual convention is that a separate copyright and license statement is a good thing to make it clear who owns the copyrights and what the licenses are. Also, not everyone receives the library from GitHub and is able to view the header on the page for various reasons.

srowen: I'm following ASF conventions here for including license info: https://www.apache.org/legal/resolved.html The license from third party libraries does need to be reproduced somewhere in distributions, and there are a bunch of ways to do it, but this seems like a fine standard to follow. 

It's already labeled in metadata, pom.xml, that is read by Github and other tools, for anything that wants something machine-readable.

What problem are you solving here ...

.reuse/dep5
andrewshadura: [REUSE](https://reuse.software/) can read this file, as can a lot of license evaluation tools that can parse the Debian copyright format (e.g. Fossology and others). Having the detailed copyright and license data for the whole project is beneficial for the users of the library who need to comply with the licenses.

Maintaining this file can be helped by e.g. Debian who will be able to pull and reuse it and later contribute updates back. Alternatively, keeping it up-to-date by the project itself can give you a good overview of the licensing situation.

(I’m a Debian Developer so I can help with the Debian bits of the above.)

srowen: Hm OK. I doubt the project will change much at all, so there isn't much worry about maintaining the updates. It's yet another piece of metadata, but, OK if it has some value to debian packaging. I'd merge this change without the others.

andrewshadura: Okay. I still think the other changes are useful, but it’s up to you, of course, to not include them if you choose so.

I will resubmit the PR without the other commits.

srowen:
Ironically, this fails the plugin that checks for correct license headers. In `pom.xml`, under `apache-rat-plugin`, if you add `<exclude>.reuse/*</exclude>`, it should be fine.

PR #1299, opened 2020-07-01
Update ResultParser.java

takeya:
This adds the feature to have multiple receivers in MATMSG:TO: mail formatted QR-Codes
without this, the parser will not recognise QR-Codes with multiple receivers in DoCoMo format and will just parse it as text

### DIFF
--- a/core/src/main/java/com/google/zxing/client/result/ResultParser.java
+++ b/core/src/main/java/com/google/zxing/client/result/ResultParser.java
@@ -228,7 +228,10 @@ public abstract class ResultParser {
             element = element.trim();
           }
           if (!element.isEmpty()) {
-            matches.add(element);
+            List<String> mailAdresses = element.split(','); // split the email adresses by comma 
+            for(String mail : mailAdresses){
+              matches.add(mail);
+            }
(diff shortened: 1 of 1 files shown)


srowen:
This code is shared to parse lots of fields. I'm concerned it will break other usages where a comma isn't used to separate items. I'm not sure it's 'allowed' by MATMSG anyway? not sure, it's such an old standard and not well documented.

ariffin89:
please support my devices

core/src/main/java/com/google/zxing/client/result/ResultParser.java
```diff
@@ -228,7 +228,10 @@ static String urlDecode(String encoded) {
             element = element.trim();
           }
           if (!element.isEmpty()) {
-            matches.add(element);
+            List<String> mailAdresses = element.split(','); // split the email adresses by comma 
+            for(String mail : mailAdresses){
```
mihaidi, line 232: use here forEach is more clear, in my opinion
mailAdresses.forEach(mailAddres->matches.add(mailAddres));

Issue #1312, opened 2020-09-14
MultiFinderPatternFinder is not a public class


yerar:
I don't see any reason why MultiFinderPatternFinder can not be a public class just as its parent FinderPatternFinder.

It may help other developers working on multi QR localization using ZXing as a dependency on java/android projects.

srowen:
Will handle this in #1313

PR #1313, opened 2020-09-14
Make MultiFinderPatternFinder a public class

yerar:
As comented before in #1312 , MultiFinderPatternFinder can be a public class like its parent.

### DIFF
--- a/core/src/main/java/com/google/zxing/multi/qrcode/detector/MultiFinderPatternFinder.java
+++ b/core/src/main/java/com/google/zxing/multi/qrcode/detector/MultiFinderPatternFinder.java
@@ -86,7 +86,7 @@ public final class MultiFinderPatternFinder extends FinderPatternFinder {
     }
   }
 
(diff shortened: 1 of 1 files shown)


srowen:
I'm not against it, but what's your use case? It's a bit of an implementation detail, so that's why it wasn't public.
You would need to make more than the class public to use it, I think.

yerar:
> I'm not against it, but what's your use case? It's a bit of an implementation detail, so that's why it wasn't public.
> You would need to make more than the class public to use it, I think.

I'm working on a project with some QR detection problems to be solved. I only need to know where are the Finder Patterns in the image, but I don't need all the additional information as the version, the alignment patterns or the decoded data. Once I have the Finders info, I apply custom transformations (such as cylindrical) to the QR.

So, for performance reasons, using MultiFinderPatternFinder is the way to go, instead of MultiDetector or QRCodeMultiReader that compute additional information not needed.

srowen (commented):
OK, I think you'll need the constructor on line 89 to be public too.

yerar:
Done!

Issue #1327, opened 2020-10-26
Aztec decoder does not handle FNC1 or ECI sequences


dlenski:
In Aztec's `PUNCT` mode, [the sequence `00000` encodes a `FLG(n)` code](https://en.wikipedia.org/wiki/Aztec_Code#Character_set) which can be used to insert either `FNC1` or [an ECI indication](https://en.wikipedia.org/wiki/Extended_Channel_Interpretation).

ZXing's Aztec decoder simply ignores the 5 bits which introduce this sequence:

https://github.com/zxing/zxing/blob/b1c85db64e0ef13e7d8a4c9de32bd94c76eea5d8/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java#L62-L65

… which causes it to not only ignore the FNC1 or ECI indicator, but also to completely lose synchronization due to the {3,7,11,15,19,23,27} bits which follow `FLG(n)` and indicate the value of `n` (3 bits) and/or the ECI code (4/8/…/24 bits).

I think it should be relatively straightforward to adopt the ECI handling from QR and PDF417 codes and add them to the Aztec decoder (and encoder), so that non-Latin1 characters can be reliably encoded in Aztec.

srowen:
I don't know Aztec well at all but take your word for it. Yes I'd take a pull request on this if you have time and motivation. It does sound straightforward

srowen:
Resolved by https://github.com/zxing/zxing/pull/1328

PR #1328, opened 2020-10-27
Aztec Code: decode ECI and FNC1

dlenski:
Addresses #1327.

- Decode FNC1 (as `0x1d`/escape character; same as Code128, DataMatrix, others appear to do)
- Decode ECI (modeled on PDF417 `DecodedBitStreamParser`)
- Fix Aztec Code encoder tests
- Include `ecLevel` estimation (percentage) in Aztec `DecodedResult`
- Add tests which specifically cover Aztec ECI decoding?

TODO:

- Handle Aztec ECI **en**coding?


core/src/test/java/com/google/zxing/aztec/encoder/EncoderTest.java
```diff
@@ -577,15 +577,21 @@ private static void testHighLevelEncodeString(String s, String expectedBits) {
     BitArray bits = new HighLevelEncoder(s.getBytes(StandardCharsets.ISO_8859_1)).encode();
     String receivedBits = stripSpace(bits.toString());
     assertEquals("highLevelEncode() failed for input string: " + s, stripSpace(expectedBits), receivedBits);
-    assertEquals(s, Decoder.highLevelDecode(toBooleanArray(bits)));
+    try {
+      assertEquals(s, Decoder.highLevelDecode(toBooleanArray(bits)));
+    } catch (FormatException e) {
```
dlenski, line 582: Not sure if there's another way that you'd prefer this to be handled @srowen.

srowen: Just let the method throw FormatException. We want it to fail the test if unexpected.

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
```diff
@@ -92,11 +96,15 @@ public static String highLevelDecode(boolean[] correctedBits) {
...
-  private static String getEncodedData(boolean[] correctedBits) {
+  private static String getEncodedData(boolean[] correctedBits) throws FormatException {
     int endIndex = correctedBits.length;
     Table latchTable = Table.UPPER; // table most recently latched to
     Table shiftTable = Table.UPPER; // table to use for the next read
-    StringBuilder result = new StringBuilder(20);
+
+    ByteArrayOutputStream decodedBytes = new ByteArrayOutputStream(); // buffer of decoded bytes
+    Charset encoding = StandardCharsets.ISO_8859_1; // current character encoding (set by ECI)
+    StringBuilder result = new StringBuilder(correctedBits.length >> 2); // final string result
```
srowen, line 106: Maybe just divide by 4 for clarity

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
```diff
@@ -142,12 +185,14 @@ private static String getEncodedData(boolean[] correctedBits) {
...
         } else {
-          result.append(str);
+          byte[] b = str.getBytes(StandardCharsets.US_ASCII);
+          decodedBytes.write(b, 0, b.length);
           // Go back to whatever mode we had been in
           shiftTable = latchTable;
         }
       }
     }
+    result.append(new String(decodedBytes.toByteArray(), encoding));
```
srowen, line 195: Does this need to check whether the byte array is empty? if it results in an empty string, fine

srowen (changes requested):
Encoding can be left to another change. We do probably need a test here, at least one.

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
```diff
@@ -196,13 +241,23 @@ private static String getCharacter(Table table, int code) {
     }
   }
 
+  static final class CorrectedBitsResult {
+    private final boolean[] correctBits;
+    private final String ecLevel;
```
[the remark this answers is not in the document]
dlenski: That's correct. I can make it an `int` here, but the `DecodedResult` will still expect it as a `String`

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
dlenski: :+1:

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
```diff
@@ -131,7 +139,42 @@ private static String getEncodedData(boolean[] correctedBits) {
...
+            break;
+          }
+          int n = readCode(correctedBits, index, 3);
+          index += 3;
+          switch (n) {
+            case 0:
+              result.append((char) 29);  // translate FNC1 as ASCII 29
+              break;
+            case 7:
+              throw new IllegalArgumentException("FLG(7) is reserved and illegal");
```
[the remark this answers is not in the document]
dlenski: :+1:. I was seeing `IllegalArgumentException` used in the QR decoder, and wasn't sure which was best.

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
dlenski: > Does this need to check whether the byte array is empty?

Yes, it works fine with an empty string.

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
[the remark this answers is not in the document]
srowen: Yeah I think I tended to use IAE for caller errors to APIs, but not for problems decoding barcodes. Not sure I got that 100% right but the decoders shouldn't generally throw IAE

srowen:
Looks good, thank you.

PR #1330, opened 2020-11-05
Aztec encode with ECI for non-default character sets

dlenski:
Follow-up to https://github.com/zxing/zxing/pull/1328, which added charset-aware _decoding_ to Aztec Code.

* Make `AztecWriter` and `aztec.Encoder` insert ECI according to character set hint
* Add encoder and round-trip test cases using ECI (and fix an erroneous test case that only passed previously because it wasn't really doing a charset-aware decode)
* Simplify test cases to avoid unnecessary `String`/`byte[]` and `Charset`/`String` (charset name) conversions.
* Further removal of unnecessary `Charset`/`String` (charset name) conversions…
  1. First, throughout the Aztec implementation
  2. In PDF417, QR, and DataMatrix implementations too

### DIFF
--- a/core/src/main/java/com/google/zxing/aztec/AztecWriter.java
+++ b/core/src/main/java/com/google/zxing/aztec/AztecWriter.java
@@ -24,7 +24,6 @@ import com.google.zxing.aztec.encoder.Encoder;
 import com.google.zxing.common.BitMatrix;
 
 import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.util.Map;
 
 /**
@@ -39,7 +38,7 @@ public final class AztecWriter implements Writer {
 
   @Override
   public BitMatrix encode(String contents, BarcodeFormat format, int width, int height, Map<EncodeHintType,?> hints) {
-    Charset charset = StandardCharsets.ISO_8859_1;
+    Charset charset = null; // Do not add any ECI code by default
     int eccPercent = Encoder.DEFAULT_EC_PERCENT;
     int layers = Encoder.DEFAULT_AZTEC_LAYERS;
     if (hints != null) {
@@ -62,7 +61,7 @@ public final class AztecWriter implements Writer {
     if (format != BarcodeFormat.AZTEC) {
       throw new IllegalArgumentException("Can only encode AZTEC, but got " + format);
     }
-    AztecCode aztec = Encoder.encode(contents.getBytes(charset), eccPercent, layers);
+    AztecCode aztec = Encoder.encode(contents, eccPercent, layers, charset);
     return renderResult(aztec, width, height);
   }
 
--- a/core/src/main/java/com/google/zxing/aztec/encoder/HighLevelEncoder.java
+++ b/core/src/main/java/com/google/zxing/aztec/encoder/HighLevelEncoder.java
@@ -169,7 +169,7 @@ public final class HighLevelEncoder {
   public BitArray encode() {
     State initialState = State.INITIAL_STATE;
     if (charset != null) {
-      CharacterSetECI eci = CharacterSetECI.getCharacterSetECIByName(charset.name());
+      CharacterSetECI eci = CharacterSetECI.getCharacterSetECI(charset);
       if (null == eci) {
         throw new IllegalArgumentException("No ECI code for character set " + charset.toString());
       }
--- a/core/src/main/java/com/google/zxing/aztec/encoder/HighLevelEncoder.java
+++ b/core/src/main/java/com/google/zxing/aztec/encoder/HighLevelEncoder.java
@@ -17,6 +17,9 @@
 package com.google.zxing.aztec.encoder;
 
 import com.google.zxing.common.BitArray;
+import com.google.zxing.common.CharacterSetECI;
+
+import java.nio.charset.Charset;
 
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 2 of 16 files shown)


core/src/main/java/com/google/zxing/aztec/AztecWriter.java
```diff
@@ -39,7 +38,7 @@ public BitMatrix encode(String contents, BarcodeFormat format, int width, int he
 
   @Override
   public BitMatrix encode(String contents, BarcodeFormat format, int width, int height, Map<EncodeHintType,?> hints) {
-    Charset charset = StandardCharsets.ISO_8859_1;
+    Charset charset = null; // Do not add any ECI code by default
```
dlenski, line 41: One thing that might be worth highlighting here is that this PR makes the semantics of the Aztec `Encoder`/`AztecWriter` slightly different from QR/PDF417/DataMatrix in terms of character set handling: specifying `charset == null` or no encoding hint means "do not add any ECI code; compliant decoders should treat as ISO_8859_1 because that's what the spec says."

The QR decoder in particular differs from this…

https://github.com/zxing/zxing/blob/0cf3b9be71680f50c90a71ca26ce0d33664b0dd6/core/src/main/java/com/google/zxing/qrcode/decoder/DecodedBitStreamParser.java#L235-L244

_Changing_ the behavior of the QR decoder would probably break a _ton_ of software out there which relies on ZXing's behavior of guessing the encoding, so I'm certainly not suggesting it :sweat_smile:.

If anything, it _might_ be a good idea to add the _same_ encoding-guessing behavior for PDF417/DataMatrix/Aztec, in case there are broken encoders out there, which encode UTF-8 without ECI.

core/src/main/java/com/google/zxing/aztec/encoder/HighLevelEncoder.java
```diff
@@ -148,16 +151,31 @@
...
+    this.charset = charset;
   }
 
   /**
    * @return text represented by this encoder encoded as a {@link BitArray}
    */
   public BitArray encode() {
-    Collection<State> states = Collections.singletonList(State.INITIAL_STATE);
+    State initialState = State.INITIAL_STATE;
+    if (null != charset) {
```
srowen, line 171: Nit: I usually write `foo != null` in this code base
Also 2-space indent rather than 4 below

core/src/main/java/com/google/zxing/aztec/encoder/State.java
```diff
@@ -70,6 +72,26 @@ int getBitCount() {
...
 
+  State appendFLGn(int eci) {
+    // FIXME: validate eci in [0,999999]
+    State result = shiftAndAppend(HighLevelEncoder.MODE_PUNCT, 0); // 0: FLG(n)
+    Token token = result.token;
+    int bitsAdded = 3;
+    if (eci < 0) {
+      token = token.add(0, 3); // 0: FNC1
+    } else if (eci > 999999) {
+      // FIXME: throw error
```
srowen, line 83: You can throw an error here like IAE, why not?

core/src/main/java/com/google/zxing/aztec/AztecWriter.java
srowen: Decoding is different from encoding, and that's a decoder you quote here. There, yes, I don't want to change the logic, but, it's also fine to take best guesses instead of failing.

core/src/main/java/com/google/zxing/common/StringUtils.java
```diff
@@ -29,28 +30,27 @@
...
 public final class StringUtils {
 
-  private static final String PLATFORM_DEFAULT_ENCODING = Charset.defaultCharset().name();
-  public static final String SHIFT_JIS = "SJIS";
-  public static final String GB2312 = "GB2312";
-  private static final String EUC_JP = "EUC_JP";
-  private static final String UTF8 = "UTF8";
-  private static final String ISO88591 = "ISO8859_1";
+  private static final Charset PLATFORM_DEFAULT_ENCODING = Charset.defaultCharset();
+  public static final Charset SHIFT_JIS = Charset.forName("SJIS");
```
srowen, line 34: This gets flagged by the binary compatibility check. While it's probably in practice not a field anyone uses, let's keep the existing public symbols for full compatibility and introduce new ones, even if the old ones are then superfluous.

srowen: (These were Strings originally simply because this originally had to work on J2ME, Java 1.3 or something, well before Charset. I agree Charset is better)

core/src/main/java/com/google/zxing/common/StringUtils.java
```diff
@@ -29,28 +30,27 @@
...
-   * @return name of guessed encoding; at the moment will only guess one of:
-   *  {@link #SHIFT_JIS}, {@link #UTF8}, {@link #ISO88591}, or the platform
-   *  default encoding if none of these can possibly be correct
+   * @return Charset of guessed encoding; at the moment will only guess one of:
+   *  {@link #SHIFT_JIS}, {@link StandardCharsets#UTF_8},
+   *  {@link StandardCharsets#ISO_8859_1}, or the platform default encoding if
+   *  none of these can possibly be correct
    */
-  public static String guessEncoding(byte[] bytes, Map<DecodeHintType,?> hints) {
+  public static Charset guessEncoding(byte[] bytes, Map<DecodeHintType,?> hints) {
```
srowen, line 51: Likewise I think we actually have to preserve this method for compatibility, just in case. Just make a new differently-named one and have the old one call it.

srowen (changes requested):

core/src/main/java/com/google/zxing/common/StringUtils.java
dlenski: Aha, makes sense. Will fix.

core/src/main/java/com/google/zxing/common/StringUtils.java
dlenski: Got it. :ok_hand:

core/src/main/java/com/google/zxing/aztec/AztecWriter.java
srowen: I don't know whether Aztec requires an ECI; if so and it's missing then that's reasonably an error. QR codes don't, so need to try something when not specified. Nevertheless, sure, sometimes it's useful to try to recover when an ECI really should be there but isn't, hence the attempt to guess common encodings.

dlenski: Aztec Code spec does not _require_ an ECI. Like most (all?) other 2D barcodes it's supposed to be read as ISO-8859-1 in the absence of an ECI.

> QR codes don't, so need to try something when not specified.

[As I understand it](https://www.linkedin.com/pulse/enhanced-channel-interpretation-terry-burton/), QR is also _supposed to be_ interpreted as ISO-8859-1 in the absence of ECI… but many encoders write UTF-8 without an ECI mark.

>  Nevertheless, sure, sometimes it's useful to try to recover when an ECI really should be there but isn't, hence the attempt to guess common encodings.

Yep, exactly.

srowen (approved):
Looks OK to me. I'll merge if you don't have any other changes right now.

dlenski:
> Looks OK to me. I'll merge if you don't have any other changes right now.

Great! I'll add charset guessing/autodetection in the decoder in a follow-up.

PR #1331, opened 2020-11-05
update Fragment and Activity from androidx (not deprecated anymore)

rafipanoyan:
`Fragment` and `Activity` classes used in android-integration are the deprecated ones. 

The goal of this PR is to update them to androidx libraries

### DIFF
--- a/android-integration/pom.xml
+++ b/android-integration/pom.xml
@@ -34,6 +34,55 @@
       <type>aar</type>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>androidx.arch.core</groupId>
+      <artifactId>core-common</artifactId>
+      <version>2.0.1</version>
+      <type>aar</type>
+    </dependency>
+    <dependency>
+      <groupId>androidx.lifecycle</groupId>
+      <artifactId>lifecycle-viewmodel</artifactId>
+      <version>2.2.0</version>
+      <type>aar</type>
+    </dependency>
+    <dependency>
+      <groupId>androidx.lifecycle</groupId>
+      <artifactId>lifecycle-livedata-core</artifactId>
+      <version>2.2.0</version>
+      <type>aar</type>
+    </dependency>
+    <dependency>
+      <groupId>androidx.core</groupId>
+      <artifactId>core</artifactId>
+      <version>1.1.0</version>
+      <type>aar</type>
+    </dependency>
+    <dependency>
+      <groupId>androidx.lifecycle</groupId>
+      <artifactId>lifecycle-runtime</artifactId>
+      <version>2.2.0</version>
+      <type>aar</type>
+    </dependency>
+    <dependency>
+      <groupId>androidx.lifecycle</groupId>
+      <artifactId>lifecycle-viewmodel-savedstate</artifactId>
+      <version>2.2.0</version>
+      <type>aar</type>
+    </dependency>
+    <dependency>
+      <groupId>androidx.annotation</groupId>
+      <artifactId>annotation</artifactId>
+      <version>1.1.0</version>
+      <type>aar</type>
+    </dependency>
+    <dependency>
+      <groupId>androidx.collection</groupId>
+      <artifactId>collection</artifactId>
+      <version>1.1.0</version>
+      <type>aar</type>
+    </dependency>
+    
   </dependencies>
 
   <repositories>
--- a/android-integration/pom.xml
+++ b/android-integration/pom.xml
@@ -14,7 +14,8 @@
  See the License for the specific language governing permissions and
  limitations under the License.
  -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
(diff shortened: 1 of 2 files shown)


srowen:
I don't want to break compatibility with older versions of Android, because older apps may still target them and still want to use this integration. Is there a later version of Android where this doesn't work at all, or merely deprecated?

rafipanoyan:
The thing is that since androidx has shipped, every applications use androix.fragment.app.Fragment or androidx.activity.Activity. All these applications cannot use this IntentIntegrators.

Some forks exist to add support for androidx classes but it would be great if the official library suported them. Maybe we could add some new constructors ?

srowen:
Is this an official library from Android or something else? does this code as is not work on certain newer Android versions?
Adding additional support/constructors isn't out of the question yes.

srowen:
Also this artifact doesn't seem to be in Maven?
https://travis-ci.com/github/zxing/zxing/jobs/428818618
https://search.maven.org/search?q=androidx.activity

rafipanoyan:
This is an official library from Google. It's the new nomenclature of the previous android.support library.  

The code as in 3.4.1 won't permit to use IntentIntegrator with the recommanded androidx classes. To make it work, the use of the deprecated Android and Fragment classes would be mandatory, but all of the existing Google libraries wont work with these. 

Concerning the artifact, you are right I forgot that they are hosted in the [Google Maven repository](https://developer.android.com/studio/build/dependencies#google-maven). 

To me, there won't be any issue to just update the current constructors since it's pretty sure no new application would start using the deprecated Fragment/Activity classes; it would mean to be incompatible with every current Google libraries.
But if adding new constructors is an option, I could just do that.

Also, Travis CI failed because androidx libraries are distributed in aar form, not jar. It is ok to use this form ?

srowen:
I'm still not clear if you're saying the old classes are merely deprecated, or not present at all, in the latest Android? 
I'm not sure how to use aar files in the build; maybe you need to `<type>aar</type>` in the dependencies?

rafipanoyan:
The old classes are still present in recent Android versions, but are not used anymore.

When creating a screen as a developer, we have to choose to inherit our custom Activities and Fragment from either : 
- android.app.Activity
- androidx.activity.Activity

Inheriting from android.app.Activity is not an option since it's deprecated and no Google's libraries use it. We have to inherit from androidx.activity.Activity, and this remove the option to use Zxing android-integration.

Is this more clear ?

srowen:
I see, so it 'works' but the calling app would have to itself use the deprecated classes.
I think it's OK to add an additional class or methods to support the new interfaces.
It would just require figuring out how to include the dependency in aar form, and it should be 'provided' scope.

rafipanoyan:
Exactly, that's why I was pretty confident about removing support for the deprecated classes since I really think nobody still uses them but a new method is doable.

I did not understand the 'provided' scope comment, I'm not experienced with the Maven build tool, what does it mean ?

srowen:
Oh just `<scope>provided</scope>` in the dependency. That way it is not packaged with the library... or is it not provided by Android?

rafipanoyan:
Androidx libraries are not provided by Android SDK, it's a separate dependency.

Do you know how should I set the Google Maven Repository where these libs are ?

```
<repositories>
    <repository>
      <id>google-maven-central</id>
      <name>Google Maven</name>
      <url>https://maven.google.com</url>
    </repository>
  </repositories>
```

Is it ok if I add this to <project> in the android-integration pom.xml ?

srowen:
Yes that's fine to try

rafipanoyan:
Would you know why Travis CIs job all pass except the openjdk8 ? It fails saying that the androidx.fragment dependency is not found, but the other jobs passed.

srowen:
The java 8 one is the only one that tests Android stuff

rafipanoyan:
Is it possible to disable the Enforcer rules for a particular dependency ? 

The androidx.fragment:fragment library has many sub-dependencies that, in return, have some different versions of the same library ([see here](https://travis-ci.com/github/zxing/zxing/jobs/429764568#L885)). 
I cannot use exclusions since I'm not in control of the androidx.fragment pom.xml. It would be easier for us to just disable the enforcer rule that throws Dependency Convergence errors for this particular androidx.fragment dependency.

srowen:
That's unfortunate, yeah. Are there newer versions that fix it?
If not you can copy the enforcer declaration from the parent pom and yes just add a single element to its configuration that disables it.

kyze8439690:
For somebody who is eager to use `IntentIntegrator` with androidx fragment library, use code snippet below: 

MainActivity which host MainFragment
```
override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
    // workaround for zxing library not support androidx fragment
    if (requestCode == IntentIntegrator.REQUEST_CODE) {
        mainFragment?.onActivityResult(requestCode, resultCode, data)
    } else {
        super.onActivityResult(requestCode, resultCode, data)
    }
}
```

MainFragment which use IntentIntegrator 
```
IntentIntegrator(getActivity()).initiateScan(QR_CODE_TYPES)

override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
    val result = IntentIntegrator.parseActivityResult(requestCode, resultCode, data)
    if (result != null) {
        // use zxing decoded result here
    } else {
        super.onActivityResult(requestCode, resultCode, data)
    }
}
```

PR #1333, opened 2020-11-08
PDF417 detector: try 90 and 270 rotations also

dlenski:


srowen (approved):
I think it's OK, yes. The only question maybe is, do you check 90/270 only in try harder mode? but I think it's OK.

dlenski:
> I think it's OK, yes. The only question maybe is, do you check 90/270 only in try harder mode? but I think it's OK.

That's why I made it try 180° first, since it's a substantially lower-cost operation to do 180° rotation. Can change it if you want, but I'd be inclined to leave it this way for symmetry with the behavior of the other 2D detectors.

PR #1348, opened 2021-01-15
Update codeql-analysis.yml

dingjiedanielyang:


srowen:
Why is this a useful change?

dingjiedanielyang:
@srowen , sorry for the confusion. This is not a useful change and please ignore it.  I was trying to check whether I could add some customized rules and trigger the QLCode scan.

PR #1350, opened 2021-01-19
Update dimension ratio change to new location (updates DMRE PR #323)

dehnhard:


srowen:
Hm unfortunately this causes a few DM images to fail to decode. It's not out of the question to just accept that, but I wonder if this change is essential, or can be less of a change to the constants here, and still achieve its goals (and pass tests)?

dehnhard:
Sure. You could try and find the lowest constant between 5 and 7 without test failures as a compromise between supporting DMRE and preventing possible regressions. ATM I can't say what that failure means or predict how many DMRE dimensions will be affected by that.

srowen:
FWIW, looks like using 6 instead of 7 still passes tests. I also don't know what that does to DMRE support, but I suppose it's an experiment now anyway. If you want to up the PR, cool, otherwise I can make a separate commit.

dehnhard:
Voilà. I added a commit with 6 - I like how it's possible to edit such simple changes directly in the browser.

And yes "experimental support" would be an apt description for this. I hope I get a chance to revisit this at some point during my work and do it properly.

PR #1367, opened 2021-03-08
Adding all readable element in the qrcode stream to byte streams.

kalaspuffar:
Hi Team.

This is one of the "I don't know if you want this fix" kinda situation.

We recently noticed that some encryption keys and URLs in our login client SQRL did not evaluate correctly after reading a QRCode. I wondered why so I did some investigation and realized that this library only added the bytes extracted from the raw data into byte streams. Hence, in the cases where QRCodes were compressed and they used characters or numbers, we did not get the bytes for these into our client.

For URLs that might be OK, we could look at the text string extracted instead, but for keys, we need the bytes to handle them correctly.

This fix will add all elements of the QRCode into the set of byte streams extracted from the data.

I hope this is something you could consider. At the moment, we are required to run the legacy version because we support Android clients from version 16 and up.

Best regards
Daniel

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/decoder/DecodedBitStreamParser.java
+++ b/core/src/main/java/com/google/zxing/qrcode/decoder/DecodedBitStreamParser.java
@@ -102,7 +102,7 @@ final class DecodedBitStreamParser {
             int subset = bits.readBits(4);
             int countHanzi = bits.readBits(mode.getCharacterCountBits(version));
             if (subset == GB2312_SUBSET) {
-              decodeHanziSegment(bits, result, countHanzi);
+              decodeHanziSegment(bits, result, countHanzi, byteSegments);
             }
             break;
           default:
@@ -111,16 +111,16 @@ final class DecodedBitStreamParser {
             int count = bits.readBits(mode.getCharacterCountBits(version));
             switch (mode) {
               case NUMERIC:
-                decodeNumericSegment(bits, result, count);
+                decodeNumericSegment(bits, result, count, byteSegments);
                 break;
               case ALPHANUMERIC:
-                decodeAlphanumericSegment(bits, result, count, fc1InEffect);
+                decodeAlphanumericSegment(bits, result, count, fc1InEffect, byteSegments);
                 break;
               case BYTE:
                 decodeByteSegment(bits, result, count, currentCharacterSetECI, byteSegments, hints);
                 break;
               case KANJI:
-                decodeKanjiSegment(bits, result, count);
+                decodeKanjiSegment(bits, result, count, byteSegments);
                 break;
               default:
                 throw FormatException.getFormatInstance();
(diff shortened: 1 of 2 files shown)


srowen (commented):
I'm not against returning raw byte segments in more places, but, can you elaborate your use case? if you are choosing to compress the bytes in the byte segments, you'd get back compressed bytes and have to decompress them, yes, but, that's app-specific.

I also didn't understand the changing it into strings of hex digits or something?
This feels somehow very specific to your app, not QR codes in general, but not quite sure yet.

core/src/main/java/com/google/zxing/qrcode/decoder/DecodedBitStreamParser.java
```diff
@@ -111,16 +111,16 @@ static DecoderResult decode(byte[] bytes,
             int count = bits.readBits(mode.getCharacterCountBits(version));
             switch (mode) {
               case NUMERIC:
-                decodeNumericSegment(bits, result, count);
+                decodeNumericSegment(bits, result, count, byteSegments);
                 break;
               case ALPHANUMERIC:
-                decodeAlphanumericSegment(bits, result, count, fc1InEffect);
+                decodeAlphanumericSegment(bits, result, count, fc1InEffect, byteSegments);
```
srowen, line 117: OK more specifically - there are no 'bytes' encoded in anything but byte mode. There are streams of bits, but I don't see much use case in returning them. Their representation is entirely specific to the QR code, not any user payload. Any use of these modes to encode arbitrary bytes would be wrong. So if you're putting raw compressed bytes into the QR code without using byte mode, that's the problem.

kalaspuffar:
Hi @srowen 

You might be correct that this is only important to my application, as I said when I filed the bug report. "This might not be something you want."

In my application, I need all the decoded bytes of the stream independent of the encoding method. When you work with crypto, you can't lose a bit anywhere, then the key or keying material would not be correct.

And even if I think that everyone on the internet should use another authentication method than Username and Password, I still need to support it. The same goes for the encoding of QRCodes. I can't tell all server and client implementations that they NEED to do the right thing and implement all the bytes as byte streams.

Thank you for your time and for considering this PR.

If you are curious, here is my application:
https://github.com/kalaspuffar/secure-quick-reliable-login

Best regards
Daniel

srowen:
I think the issue here is that QR codes don't _really_ encode bytes. They encode text. There is a byte mode, meant for encoding general text as ISO-8859-1 but can be used for other encodings. And that could be used to encode arbitrary bytes. But any normal decoder has no idea of those semantics and would try to render it as the (gibberish) text it represents.

If you are trying to do something with arbitrary QR codes (?) then you have to treat their content as text. You can re-encode that as bytes to operate on the as bytes.

Another thing I want to emphasize is that in your example, in no sense is the byte 65 present in the input logically, nor present inside the QR code. "A" is present in the second segment. You can choose to interpret that as value 65 according to ASCII, but, that's nothing to do with what's in the bits in the QR code in this case. 

Pushing that interpretation inside the decoder doesn't make a lot of sense to me, since it is highly specific to what you're trying to do, it's ambiguous (why ASCII? or UTF-8? etc) , and because you as the caller of the library can do that perfectly well already.

PR #1369, opened 2021-03-17
Update DMRE versions to ISO/IEC 21471:2020

gitlost:
Updates the DMRE versions used by PR #323 to ISO/IEC 21471:2020. (See also ZXing-C++ https://github.com/nu-book/zxing-cpp/pull/201)

Also backs out the changes https://github.com/zxing/zxing/commit/e6713ae554c44fb158e49e7139a5441165db71af which should be covered by the adjustment to the row index in `BitMatrixParser::readModule()`, an addition given in the ISO 21471:2020 E.1 Symbol placement character module() routine.

The test files added are "pure" ones created using [Zint](https://sourceforge.net/projects/zint/). One fails when rotated 90 degrees for some reason. Also the photo examples from the [EuroData Council](https://e-d-c.info/en/projects/dmre-en.html) site added to ZXing-C++ have been left out as they fail for some reason. (I've a feeling the transition tracing code used to determine the dimensions wanders off course.)


srowen (approved):
OK I trust your work here, looks good.

oehhar:
Thank you all for the action. I have put this reference to [www.dmre.info](www.dmre.info).
If there is a more user-friendly link, I would like to add it there.
I appreciate,
Harald

PR #1371, opened 2021-03-29
inverted barcode support

dehnhard:
Adds a hint (POSITIVE_AND_NEGATIVE) to automatically and additionally scan for inverted barcodes.

- A similar implementation (without a hint) has been done in the ios port (https://github.com/zxingify/zxingify-objc/pull/499).
- A test is provided with a datamatrix code, but the implementation works for all barcode types.


core/src/main/java/com/google/zxing/BinaryBitmap.java
```diff
@@ -138,6 +138,16 @@ public BinaryBitmap rotateCounterClockwise45() {
...
   }
 
+  /**
+   * Returns a new object with negative image data.
+   *
+   * @return An inverted version of this object.
+   */
+  public BinaryBitmap invert() {
+    LuminanceSource newSource = binarizer.getLuminanceSource().invert();
+    return new BinaryBitmap(binarizer.createBinarizer(newSource));
```
srowen, line 148: I think it might be more efficient to just add a flip() method for the whole BitMatrix, and call it here? This is sound too as it reuses the (unused!) LuminanceSource.invert() method but has to binarize the result again

srowen (changes requested):
Already OK but looks like there is a test failure - does it depend on another PR?

core/src/main/java/com/google/zxing/BinaryBitmap.java
dehnhard: Funny that you point that out. Inverting the BitMatrix itself was actually my first implementation.

I switched to the existing invert method because I try to reuse existing code as much as possible, when contributing to a foreign code base. I'll gladly switch back to the other implementation.

The test failure indicates a small leftover of https://github.com/zxing/zxing/pull/1372 when separating the code of both PRs.
I'll update the PR soon.

srowen: OK I may eventually just convert this to flipping the bits directly, but I don't mind as a first pass.

core/src/main/java/com/google/zxing/DecodeHintType.java
```diff
@@ -98,6 +98,12 @@
    */
   ALLOWED_EAN_EXTENSIONS(int[].class),
 
+  /**
+   * If true, also tries to decode as negative image. All configured decoders are simply called a
```
srowen, line 102: Can we call this 'inverted' in the docs and name for consistency? no big deal.

core/src/test/java/com/google/zxing/common/AbstractBlackBoxTestCase.java
```diff
@@ -92,6 +93,10 @@ protected final void addTest(int mustPassCount, int tryHarderCount, float rotati
     addTest(mustPassCount, tryHarderCount, 0, 0, rotation);
   }
 
+  protected void addHint(DecodeHintType hint) {
```
srowen, line 96: I'm worried it modifies a hints object reused by all tests, so one test calling this affects the next, but, looks like it's only used where it's global to all tests in a suite object.

core/src/test/java/com/google/zxing/inverted/NoHintInvertedDataMatrixBlackBoxTestCase.java
```diff
@@ -0,0 +1,34 @@
...
+package com.google.zxing.inverted;
+
+import com.google.zxing.common.AbstractNegativeBlackBoxTestCase;
+
+/**
+ * Without hint inverted barcodes should not be found.
+ */
+public final class NoHintInvertedDataMatrixBlackBoxTestCase extends AbstractNegativeBlackBoxTestCase {
+
+  public NoHintInvertedDataMatrixBlackBoxTestCase() {
```
srowen, line 26: I'd merge this with the test suite above (or just drop it, honestly - the not-inverted case is already well tested) If so then addHint has to change a little bit.

srowen (commented):
I think it's pretty OK - see some comments.
If I have follow-ons I can change it myself.

There is functionality in the Barcode Scanner app to invert an image that could leverage this hint now, but, probably not worth modifying.

core/src/test/java/com/google/zxing/common/AbstractBlackBoxTestCase.java
dehnhard: I chose this method because it looked liked the way with the minimal impact to pass a hint to the decoder.

The hint object is created deep down in the decode method and fitted with several test related values. What I'd like to have used is a parameter to pass an arbitrary hint to the addTest method. I think that would be helpful for other tests, too. I just didn't felt I understand the logic of the test class sufficently enough to introduce that myself.

core/src/test/java/com/google/zxing/inverted/NoHintInvertedDataMatrixBlackBoxTestCase.java
dehnhard: This test actually tests if the same inverted barcode used in the other test is NOT scanned, when the hint is NOT set. It could be called the negative inverted test :-)

core/src/main/java/com/google/zxing/DecodeHintType.java
dehnhard: BinaryBitmap.invert() is actually the old method. I added the new method BitMatrix.flip() and changed the method call, but forgot to remove BinaryBitmap.invert(). I'll do that now.

I renamed negative to inverted. I called the hint ALSO_INVERTED, which should reflect the meaning well (but I'm open for better ideas) while POSITIVE_AND_INVERTED sounds ugly to me.

srowen:
OK close enough. I may fiddle with the change a little internally later.

PR #1372, opened 2021-03-29
Symbology Identifier support

dehnhard:
This PR proposes support for symbology identifiers in zxing.

- The symbology identifier is not prepended to the barcode content, as this would be quite a major breaking change for an established library. Instead the decoder now returns new metadata field called SYMBOLOGY_IDENTIFIER with the three symbology characters (']' + identifier for the barcode type + type specific modifier number, e.g. ']d1').

- The symbology modifiers are different for every barcode type. For some types this implementation just returns a basic default value.


oehhar:
This is IMHO a great improvement, thank you !

srowen:
I'm not against it but why this particular format, and what does the identifier mean? is there a standard that defines these numbers?

oehhar:
Dear Sean,
thank you for asking. This format is the standard format defined by ISO/IEC15424. The older document is "AIM Symbology ID Specification". Each symbology standard has a chapter how this symbology ID is returned.

GS1 is strictly not decodable without that.

The leading FNC1 character is expressed by a change in the symbology ID:

- for Code128 : from ]C0 to ]C1
- for Datamatrix: from ]d1 to ]d2

And ECI is not decodable without that neither, as its presence is flagged by the symbology ID to.

Please send me a private E-Mail, if you want more information.

Thank you and best regards,
Harald

core/src/main/java/com/google/zxing/datamatrix/decoder/DecodedBitStreamParser.java
```diff
@@ -117,15 +126,34 @@ static DecoderResult decode(byte[] bytes) throws FormatException {
...
   }
 
   /**
    * See ISO 16022:2006, 5.2.3 and Annex C, Table C.2
    */
   private static Mode decodeAsciiSegment(BitSource bits,
                                          StringBuilder result,
-                                         StringBuilder resultTrailer) throws FormatException {
+                                         StringBuilder resultTrailer,
+                                         Map fnc1positions) throws FormatException {
```
srowen, line 156: This Map can/should have generic types, no?

core/src/main/java/com/google/zxing/datamatrix/decoder/DecodedBitStreamParser.java
```diff
@@ -117,15 +126,34 @@ static DecoderResult decode(byte[] bytes) throws FormatException {
     if (resultTrailer.length() > 0) {
       result.append(resultTrailer);
     }
-    return new DecoderResult(bytes, result.toString(), byteSegments.isEmpty() ? null : byteSegments, null);
+    if (isECIencoded) {
+      if (fnc1Positions.containsKey(0) || fnc1Positions.containsKey(4)) {
+        symbologyModifier = 5;
```
srowen, line 131: If you have a link explaining these numbers, that would be good to include in a comment

core/src/main/java/com/google/zxing/maxicode/decoder/DecodedBitStreamParser.java
```diff
@@ -100,7 +100,7 @@ static DecoderResult decode(byte[] bytes, int mode) {
         result.append(getMessage(bytes, 1, 77));
         break;
     }
-    return new DecoderResult(bytes, result.toString(), null, String.valueOf(mode));
+    return new DecoderResult(bytes, result.toString(), null, String.valueOf(mode), 0);
```
srowen, line 103: If 0 is a default value, how about just overloading the constructor with a version that doesn't take this input?

core/src/main/java/com/google/zxing/oned/Code128Reader.java
```diff
@@ -238,6 +239,8 @@ public Result decodeRow(int rowNumber, BitArray row, Map<DecodeHintType,?> hints
 
     boolean convertFNC1 = hints != null && hints.containsKey(DecodeHintType.ASSUME_GS1);
 
+    String symbology_modifier = "0";
```
srowen, line 242: Nit: symbologyModifier

srowen (changes requested):

core/src/main/java/com/google/zxing/maxicode/decoder/DecodedBitStreamParser.java
dehnhard: Sure, I'll add that. 
I actually would have used a parameter default if Java had support for that. Overloading is a bit of a pain.

core/src/main/java/com/google/zxing/datamatrix/decoder/DecodedBitStreamParser.java
dehnhard: Good advice.
I'm aware Java has generic types, but it's hard to change habits that have been formed in many years without them and rarely getting to code Java nowadays.

core/src/main/java/com/google/zxing/datamatrix/decoder/DecodedBitStreamParser.java
dehnhard: There are links, but they all have caveats:
- The standard ISO/IEC15424 is not publicly available (ISO paywall).
- The GS1 Specification https://www.gs1.org/docs/barcodes/GS1_General_Specifications.pdf has a list in section 5.1.3, but only lists the GS1 specific ones.
- https://honeywellaidc.force.com/supportppr/s/article/List-of-barcode-symbology-AIM-Identifiers has a more complete list, but is actually part of the documentation of a specific hardware barcode scanner product.

dehnhard:
Thanks @oehhar for pointing out the relevant standard (at least I have to presume that the logic I implemented for a client are matching that standard, as I haven't read it). 
But I have to add that neither that ISO standard nor the GS1 specification mandate that barcode scanners provide a symbology identifier (the ZXing library is basically a software based barcode scanner):
- "This identification is achieved by the addition of an **optional feature** to readers enabling the reader to prefix a standard string of characters to data messages.", taken from the Introduction of ISO/IEC 15424:2008 (https://www.iso.org/obp/ui/#iso:std:iso-iec:15424:ed-2:v1:en)
- https://www.gs1.org/docs/barcodes/GS1_General_Specifications.pdf, section 3.5.1: "Some
scanners have the **optional feature** of being able to transmit a symbology identifier. ... If used, the symbology identifier is transmitted as a prefix to the data message."

Having said, it is true that the barcode scanner is conceptually and technically the optimal place to determine and provide the symbology identifier. Side note: Thats also true for ECI decoding, which is actually implemented for some barcode types (like Aztec) but not for others (e.g. Datamatrix).


@srowen In my understanding that this symbology identifier is widely used as an intermediate representation for barcode content when dealing with different barcode types and it's implementation would prevent all of them to "reinvent the wheel" by implementing symbology determination. This PR is based on work for a client who needed exactly that,

I also noticed that the barcode readers are treating the FNC1 character (which is essential for symbology determination) very differently. I like to propose some kind of unification for zxing, but I will open a seperate issue for that discussion.

oehhar:
Yes, all this is a mess. And if you add ECI it is getting more messy.

ECI: tell the used codepage (example: UTF-8).
You know perhaps, that there are printing projects which do not use kanji compression in QR-Code, as Zxing incorrectly decodes it as kanji. It is only a compaction mode, not an information that kanji is used.
Thats what ECI is for. The free zint project has done considerable effort here.

To clarify the mess, a new ISO/IEC work item "decoder interface" was started this year.
And ISO/IEC 15424 is also open for revision.

If anybody is interested in the discussion, please drop me a private E-Mail (harald.oehlmann@e-d-c.info).

Thank you for all the great work,
Harald

srowen:
How would you treat the FNC1 char differently? it may need to be standardized, yes. It's not a printing char but I recall there are situations where it needs to be rendered as a string anyway.

What's the kanji issue? it's possible someone encodes kanji in Shift_JIS in byte mode, that's valid.

oehhar:
Dear Srowen,

thank you for the questions. About FNC1:
- an FNC1 at the first position changes the Symbology ID and is not reflected in data
- if there was an FNC1 in the first position (well, there might be an ECI before), then any following FNC1 is transmitted as ASCII GS (decimal code 29). Any other use of FNC1 is invalid.

About kanji: if one wants to encode Kanji, the ECI protocol is mandatory (otherwise, it is not possible to encode Kanji).
The process is as follows (look to the zint project for practical examples):
   -    So compose the message with a suitable codepage (examples: UTF-8, Big5)
   -   Encode the ECI corresponding to the Codepage
   -   Represent the message in the QR code. This may or may not use Kanji comaction, that is independent to the fact, that Kanji is ment.

The decoding works as follws:

   -   Due to the ECI, the Symbology ID is changed.
   -   Then ECI protocol is active and the ECI information and transformed raw data is transmitted

The document mentioned by me aims to add the capability to the decoder to directly output utf-8 (and not leave the decoding to the application).

Thank you again for the question,
Harald

srowen:
Right, which FNC1 behavior doesn't match that?

I understand ECI segments. It still tries to auto-detect character encoding without one, because lots of real QR codes don't specify one. But yes it reads and uses the one defined by ECI if present. What's the problem?

oehhar:
Sean,
thank you for all the effort. Zxing is great !

The issue is more in the principle that most people don't use ECI's and auto-detection fails.
The zint projects and friends have added a switch to zint to not use kanji compaction, as it was stated that zint takes this as a hint to transform data and binary data got corrupted. Eventually this is a tale, I don't know.

I jumped into the discussion without knowing code or even exact behaviour of zxing.
I just want to say, that without Symbology IDs, nor GS1 (FNC1), nor ECI may be decoded in the standard way.

Thank you,
Harald

core/src/main/java/com/google/zxing/maxicode/decoder/DecodedBitStreamParser.java
srowen: You can leave this as is or just add another overload, whatever makes more sense to you

dehnhard:
Took me a while to find time to write it down, but now I opened a new issue about FNC1 conversions: https://github.com/zxing/zxing/issues/1373

core/src/main/java/com/google/zxing/maxicode/decoder/DecodedBitStreamParser.java
dehnhard: Both variants are fine to me. 
As you leave it to me, I'll opt for keeping it as it is. Feels clearer to me where to modify the default in the BitStreamParsers in the case of adding a more detailed detection in the future.

dehnhard:
@srowen I updated the PR as discussed.

There is a Travis CI check which seems to dislike a method signature change comparing to the latest release:
```
[INFO] Comparing to version: 3.4.1
338[ERROR] 7004: com.google.zxing.common.DecoderResult: In method 'public DecoderResult(byte[], java.lang.String, java.util.List, java.lang.String)' the number of arguments has changed
```
That change is an intentional part of the implementation (and not a public interface).

srowen:
Right, though, it's not a private class. Just in case some impl uses the class, I'd add the overloaded constructor to keep the binary signature identical. It's worth one more constructor.

dehnhard:
> Right, though, it's not a private class. Just in case some impl uses the class, I'd add the overloaded constructor to keep the binary signature identical. It's worth one more constructor.

Makes sense.
I did that, although because of the existing overloaded parameters I had to add two more contructors. Is  that ok? If not, please just tell me how you want it to be.

core/src/main/java/com/google/zxing/datamatrix/decoder/DecodedBitStreamParser.java
```diff
@@ -117,15 +126,34 @@ static DecoderResult decode(byte[] bytes) throws FormatException {
     if (resultTrailer.length() > 0) {
       result.append(resultTrailer);
     }
-    return new DecoderResult(bytes, result.toString(), byteSegments.isEmpty() ? null : byteSegments, null);
+    if (isECIencoded) { // Examples for this numbers can be found in this documentation of a hardware barcode scanner: https://honeywellaidc.force.com/supportppr/s/article/List-of-barcode-symbology-AIM-Identifiers
```
srowen, line 129: Nit: lets wrap this line. I think it fails style checks

core/src/main/java/com/google/zxing/oned/Code128Reader.java
```diff
@@ -238,6 +239,8 @@ public Result decodeRow(int rowNumber, BitArray row, Map<DecodeHintType,?> hints
 
     boolean convertFNC1 = hints != null && hints.containsKey(DecodeHintType.ASSUME_GS1);
 
+    String symbologyModifier = "0";
```
srowen, line 242: Should this just be an int?

core/src/main/java/com/google/zxing/datamatrix/decoder/DecodedBitStreamParser.java
```diff
@@ -348,6 +375,7 @@ private static void decodeTextSegment(BitSource bits, StringBuilder result) thro
             } else {
               switch (cValue) {
                 case 27: // FNC1
+                  fnc1positions.put(result.length(), null);
```
srowen, line 378: Are the values ever anything except null? if it is used as a Set, use a Set<Integer>

core/src/main/java/com/google/zxing/maxicode/decoder/DecodedBitStreamParser.java
srowen: OK can this use the same constructor now rather than add the default value?

srowen (changes requested):

core/src/main/java/com/google/zxing/oned/Code128Reader.java
dehnhard: Aggreed for code consistency. Technical no difference, it's concatenated to a String anyway.

core/src/main/java/com/google/zxing/datamatrix/decoder/DecodedBitStreamParser.java
dehnhard: Ok. Just didn't remember Set also has a hash implementation, with hashing being important here for speed and space.

dehnhard:
I hope we are done now.

srowen (approved):

PR #1379, opened 2021-04-08
PDF417 Macro: fileId numeric decode; allow Macro only symbol

gitlost:
Decodes the Macro PDF417 "fileId" codewords as 0-899 numbers, each 0-filled to width 3. See https://github.com/nu-book/zxing-cpp/pull/219.

As mentioned there not doing any base 900 maths based on the example in ISO/IEC 15438:2015 Annex H.6, where (100, 200, 300) is transmitted as "100200300". Also as discussed there in the code review the old decoding as text can lose info if numeric codewords convert to shifts/latches etc.

Checks that `fileId` is not empty (mandatory field, Annex H.2).

Allows a symbol to consist solely of a Macro (i.e. checks for non-existence of `fileId` on result empty check), as stated in Annex H.2 note.

Unit test additions.

### DIFF
--- a/core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
+++ b/core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
@@ -18,6 +18,7 @@ package com.google.zxing.pdf417.decoder;
 
 import com.google.zxing.FormatException;
 import com.google.zxing.pdf417.PDF417ResultMetadata;
+import com.google.zxing.common.DecoderResult;
 import org.junit.Assert;
 import org.junit.Test;
 
@@ -39,7 +40,7 @@ public class PDF417DecoderTestCase extends Assert {
     DecodedBitStreamParser.decodeMacroBlock(sampleCodes, 2, resultMetadata);
 
     assertEquals(0, resultMetadata.getSegmentIndex());
-    assertEquals("ARBX", resultMetadata.getFileId());
+    assertEquals("017053", resultMetadata.getFileId());
     assertFalse(resultMetadata.isLastSegment());
     assertEquals(4, resultMetadata.getSegmentCount());
     assertEquals("CEN BE", resultMetadata.getSender());
@@ -66,7 +67,7 @@ public class PDF417DecoderTestCase extends Assert {
     DecodedBitStreamParser.decodeMacroBlock(sampleCodes, 2, resultMetadata);
 
     assertEquals(3, resultMetadata.getSegmentIndex());
-    assertEquals("ARBX", resultMetadata.getFileId());
+    assertEquals("017053", resultMetadata.getFileId());
     assertTrue(resultMetadata.isLastSegment());
     assertEquals(4, resultMetadata.getSegmentCount());
     assertNull(resultMetadata.getAddressee());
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
```diff
@@ -89,7 +116,7 @@ public void testSampleWithFilename() throws FormatException {
     DecodedBitStreamParser.decodeMacroBlock(sampleCodes, 3, resultMetadata);
 
     assertEquals(0, resultMetadata.getSegmentIndex());
-    assertEquals("AAIMAVC ", resultMetadata.getFileId());
+    assertEquals("000252021086", resultMetadata.getFileId());
```
srowen, line 119: For someone that doesn't know the details - can you explain why this result changes? was it definitely wrong?

gitlost: Hi, no, it's just how the codewords `0, 252, 21, 86` are interpreted. It's the same info in this case, and is what one would expect if one used e.g. TEC-IT and typed in "AAIMAVC " as the file ID. However the codewords e.g. `0, 252, 21, 836` would also produce "AAIMAVC ", (86 = space + pad, 836 = shift to lower + space), so info is lost and file IDs become ambiguous.

gitlost: The most compelling reason for me is that the standard says that file ID codewords are to be interpreted this way (Annex H.2.2): "Each codeword of the series is transmitted as the 3-digit ASCII representation of its decimal value." Text compaction is not mentioned.

The example in Annex H.6 says that the codeword series `100, 200, 300` should be transmitted as "100200300". 

The example in H.4 (`StandardSample1` and `StandardSample2`) gives "File ID = 17(base 900) 53(base 900)" and the codewords as `17,53`.

I don't know where the file Id used in `SampleWithFilename` and `SampleWithNumericValues` came from, but the fact that it ends with a space when decoded using text compaction is suspicious!

So yes `getFileId` will be returning the codewords, 0-filled, but this is what the standard says.

I obviously confused things with mentioning TEC-IT and info loss, but info loss is a functional reason for using the standard scheme, as text compaction can lead to two different file IDs being stored as the same text compacted string (as per the example given above), so if e.g. 2 symbol sequences with different file IDs produced by 2 different generators, TEC-IT (encoding "AAIMAVC ") and a standards-compliant one (encoding "000252021836"), were both being processed they would clash. Similarly 2 sequences "000252021083" and "000252021836" generated by one standards-compliant generator. Unlikely maybe, but possible!

srowen: OK, I don't have the standard, but if that's what it says, then I agree. I don't have background on the current behavior. If you're sure this is the 'expected' semantics then I'm on board.

gitlost: Well I wouldn't say I'm sure or know what the expected semantics are but this way makes more sense to me...

gitlost: I notice [IDAutomation](https://www.bcgen.com/pdf417-barcode-creator.html) insists on a numeric file ID and encodes it according to the standard (though only seems to allow 0-899 i.e. a single codeword). Note that 899 is an empty string if decoded using text compaction (as are a number of other "8xx" numbers, as they decode to shift/latch pairs).

srowen:
OK. Basically I even have a copy of the standard, just don't know it or want to read it.
I tend to trust people that seem to know the right behavior well, so I'm inclined to merge this.
If anyone complains I can send them to you? :)

gitlost:
Ha as long as they form an orderly socially-distanced queue...

gitlost:
I should add that if you feel this could cause back-compatibility issues for people, then I'm fine with the PR being closed unmerged.

PR #1390, opened 2021-05-16
Introduce BinaryBitmap parent, AbstractBinaryBitmap

Darkyenus:
This change should help with custom BinaryBitmap implementations, which are now awkward to create.
This way was chosen instead of making the BinaryBitmap into an interface to minimize breakage
for end users, as I assume that there are more people using BinaryBitmap than implementing custom Readers.
Making the BinaryBitmap non-final is also an option, but custom subclasses would be fragile with regards
to future BinaryBitmap changes.

This also splits isRotateSupported into two variants, for 90 and 45 degrees, to make it easier to implement
rotation support properly, given that 45 degree rotation is not used by built-in Readers, but is required
to leverage the rotation support.

*This contribution is my original work and I license the work to the project under the project's open source license.*

### DIFF
--- a/core/src/main/java/com/google/zxing/AbstractBinaryBitmap.java
+++ b/core/src/main/java/com/google/zxing/AbstractBinaryBitmap.java
@@ -1,0 +1,96 @@
+package com.google.zxing;
+
+import com.google.zxing.common.BitArray;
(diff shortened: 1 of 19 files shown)


Darkyenus:
After some further experimentation, this may not be needed for my use case after all.

PR #1396, opened 2021-06-05
Fix symbology identifier tests

makiuchi-d:
I found that some metadata tests did not work because the *.metadata.txt filename did not match to the png file.
First, in this PR, I renamed them to correct names.

And I also found that the symbology-identifier was not implemented in the ITFReader and there was no metadata test for the RSS14.
I implemented it, and added a metadata test file.

In addition, I removed unused code in the Code128Reader.
CODE_FNC_2 (value=97) is not available in the CODE_C (value=97 means "97"; see [line 458-462 of Code128Reader.java](https://github.com/makiuchi-d/zxing/blob/fix-symbology-identifier-tests/core/src/main/java/com/google/zxing/oned/Code128Reader.java#L458-L462)).


core/src/main/java/com/google/zxing/oned/Code128Reader.java
```diff
@@ -482,9 +482,6 @@ public Result decodeRow(int rowNumber, BitArray row, Map<DecodeHintType,?> hints
                   }
                 }
                 break;
-              case CODE_FNC_2:
```
srowen, line 485: I believe you, but why is this definitely unused?

makiuchi-d: This `switch(code)` is in the `else` block from line 467, and corresponding `if` is at line 458.

The line 458 is:
```
if (code < 100) {
```

Therefore, the `code` in this `switch` expression is greater than or equal to 100.
And `CODE_FNC_2` is 97 (defined at line 160), so there is no chance to match to this `case`.

srowen: OK you're right. The static analyzer agrees. @dehnhard you added this in #1372 - do you object?

dehnhard: No objection from me. Looks like CODE_C is numeric only and can't contain FNC2 (but a leading FNC1). So, no harm done, it is removable dead code.
( and good to see improvements coming in on the symbology identifier code )

PR #1410, opened 2021-07-26
Fix typo in ResultMetadataType.java

eltociear:
identifer -> identifier


srowen:
Ok, do you see any more? If you have a spell check going that's great.

PR #1411, opened 2021-07-27
Introduce hint to force specific code set for Code-128 encoding

gschwaer:
Hi,
This PR enables forcing the code set for the Code128 encoder.
This feature is important when trying to exactly recreate a barcode that has been scanned (see TheLastProject/Catima#244).
The problem and the change are described on [this post on the Google Group](https://groups.google.com/g/zxing/c/Pi3XasIqQi8/m/t-_XIY92FgAJ). I ported and tested the change for the master branch.
Thanks for considering this PR.


core/src/main/java/com/google/zxing/EncodeHintType.java
```diff
@@ -115,4 +115,9 @@
    * {@link String } value).
    */
   GS1_FORMAT,
+
+  /**
+   * Forces which encoding will be used. Introduced for Code-128 code sets (Type {@link String}).
```
srowen, line 120: You could say here that it's currently only used by the Code 128 implementation, and that the possible values are A B and C.

core/src/test/java/com/google/zxing/oned/Code128WriterTestCase.java
```diff
@@ -151,4 +155,86 @@ private void testEncode(String toEncode, String expected) throws Exception {
     String actualRoundtripResultText = rtResult.getText();
     assertEquals(toEncode, actualRoundtripResultText);
   }
+
+  @Test
```
srowen, line 159: For these, you can write `@Test(expected = IllegalArgumentException.class)` I'm not worried about the message

core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -79,13 +97,33 @@
...
+        case CODE_CODE_A:
+          if (c > 95) {
+            // the last 32 ASCII characters are not present in code set A
+            throw new IllegalArgumentException("Bad character in input for forced code set A: " + c);
+          }
+          break;
+        case CODE_CODE_B:
+          if (c < 32) {
+            // the first 32 ASCII characters are not present in code set B
+            throw new IllegalArgumentException("Bad character in input for forced code set B: " + c);
```
srowen, line 117: For this case, maybe others, do you want to include the ASCII value of the char? c is a `char` and it will be unprintable in this case

core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -79,13 +97,33 @@
...
+          }
+          break;
+        case CODE_CODE_C:
+          if (c < 48 || c > 57 || c == ESCAPE_FNC_2 || c == ESCAPE_FNC_3 || c == ESCAPE_FNC_4) {
```
srowen, line 121: I might be misreading this but doesn't `c > 57` cover the rest of the cases here?

core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
```diff
@@ -33,8 +33,11 @@
 public abstract class OneDimensionalCodeWriter implements Writer {
   private static final Pattern NUMERIC = Pattern.compile("[0-9]+");
 
+  protected Map<EncodeHintType,?> hints;
```
srowen, line 36: Rather than carry this around as a field, as it isn't otherwise used I believe, just change the implementation above - the current encode() method which does not accept hints can just call the version that does accept it, passing null, and then the current implementation becomes the body of an implementation that accepts the hints. This would be slightly more consistent with other impls.

srowen (changes requested):
Looking fine, just some minor comments

core/src/main/java/com/google/zxing/EncodeHintType.java
gschwaer: I usually tend not to write "currently" in a comment, since the code often changes faster than the comments, but I get your point.

core/src/test/java/com/google/zxing/oned/Code128WriterTestCase.java
gschwaer: That's neat!

core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
gschwaer: Yes, that's the better solution. I hope I solved it better now.

core/src/main/java/com/google/zxing/EncodeHintType.java
srowen: The code really won't change much.

core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -65,12 +67,34 @@
...
+        case "A":
+          forcedCodeSet = CODE_CODE_A;
+          break;
+        case "B":
+          forcedCodeSet = CODE_CODE_B;
+          break;
+        case "C":
+          forcedCodeSet = CODE_CODE_C;
+          break;
+      }
```
srowen, line 95: No big deal but you could throw an IAE in the default branch here

core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -82,10 +106,15 @@
           break;
         default:
           if (c > 127) {
-            // support for FNC4 isn't implemented, no full Latin-1 character set available at the moment
-            throw new IllegalArgumentException("Bad character in input: " + c);
+            // no full Latin-1 character set available at the moment
+            throw new IllegalArgumentException("Bad character in input: ascii value=" + (int) c);
```
srowen, line 110: Nit: ASCII value =

core/src/main/java/com/google/zxing/oned/OneDimensionalCodeWriter.java
```diff
@@ -144,5 +144,15 @@ public int getDefaultMargin() {
...
    */
   public abstract boolean[] encode(String contents);
+
+  /**
+   * Can be overwritten if the encode requires to read the hints map. Otherwise it defaults to {@code encode}.
+   * @param contents barcode contents to encode
+   * @param hints encoding hints
+   * @return a {@code boolean[]} of horizontal pixels (false = white, true = black)
+   */
+  protected boolean[] encodeWithHints(String contents, Map<EncodeHintType,?> hints) {
```
srowen, line 154: This is OK; you're welcome to move these two encode() methods up with the other encode() methods too

gschwaer: Sorry I don't quite understand. You mean just to move the lines up to e.g. line 76 to where it's used? Or do your mean up the inheritance tree into the Writer?

srowen: Just moving the definitions. I don't know why I put this method at the end when the other encode methods are at the top. All of them including the new one could go at the top together.

gschwaer: Ok, makes sense. I also changed the name from encodeWithHints to just encode. I mainly write C, so I often forget overloading exists. ;)

srowen (approved):

PR #1430, opened 2021-09-05
small fix: valid `y` range

makiuchi-d:
`y = 0` is valid like `x`.


srowen:
I'm always uneasy that there's some subtle reason this is how it is, but I doubt it. I think I agree.

PR #1441, opened 2021-09-28
Fix decoding of FNC1 in AztecCode.

sm6xmm:
I discovered that when decoding an Aztec GS1 barcode, all <GS> characters are output at the beginning of the stream. This patch fixes that.

### DIFF
--- a/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
+++ b/core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
@@ -153,30 +153,20 @@ public final class Decoder {
           }
           int n = readCode(correctedBits, index, 3);
           index += 3;
+          //  flush bytes, FLG changes state
+          try {
+            result.append(decodedBytes.toString(encoding.name()));
+          } catch (UnsupportedEncodingException uee) {
+            throw new IllegalStateException(uee);
+          }
+          decodedBytes.reset();
           switch (n) {
             case 0:
-              //  flush bytes before outputting FNC1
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
```diff
@@ -155,6 +155,14 @@ private static String getEncodedData(boolean[] correctedBits) throws FormatExcep
           index += 3;
           switch (n) {
             case 0:
+              //  flush bytes before outputting FNC1
+              try {
```
srowen, line 159: How about just pulling this block up before the switch statement? it happens in the 0 and default case anyway, and 7 leads to an exception anyway. (and just keep 2-space indent)

sm6xmm: Yes, this makes total sense. For 7 we don't know what to do anyway, so whatever is right.

srowen (approved):

core/src/main/java/com/google/zxing/aztec/decoder/Decoder.java
kakar930: https://app.codecov.io/gh/zxing/zxing/commit/be2c5bdd883f9f42e794e4b0e7e1b30f436fc0a7

kakar930: This commit fc50fca compared to be2c5bd

PR #1442, opened 2021-09-28
Add option to output raw bitstream from CommandLineRunner.

sm6xmm:
A small patch to add a --raw option to the CommandLineRunner to get the raw bitstream from the decode of the barcode. This is useful, especially when debugging encoding issues in barcode encoders. The same data is available in the web decoder, but sometimes the command line is easier.

### DIFF
--- a/javase/src/main/java/com/google/zxing/client/j2se/DecodeWorker.java
+++ b/javase/src/main/java/com/google/zxing/client/j2se/DecodeWorker.java
@@ -165,6 +165,17 @@ final class DecodeWorker implements Callable<Integer> {
             result.getText() + "\n" +
             "Parsed result:\n" +
             parsedResult.getDisplayResult() + "\n");
+
+        if (config.outputRaw) {
+            String rawData = "";
+
+            for (byte b : result.getRawBytes()) {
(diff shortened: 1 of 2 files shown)


javase/src/main/java/com/google/zxing/client/j2se/DecodeWorker.java
```diff
@@ -165,6 +165,17 @@ private static void dumpResult(URI input, Result... results) throws IOException
             result.getText() + "\n" +
             "Parsed result:\n" +
             parsedResult.getDisplayResult() + "\n");
+
+        if (config.outputRaw) {
+            String rawData = "";
```
srowen, line 170: Total nits but can you switch to 2-space indent? and this could be a StringBuilder

sm6xmm: Sure (strange that I missed the indent, I blame not bothering to set the correct offset in my editor for such a small change). I also switch to StringBuilder, it makes sense even though it isn't critical here. (I didn't look hard enough for Python's string.join())

srowen (approved):
There is a String.join() method in Java too, but not worth using here. This looks fine

PR #1445, opened 2021-10-01
Added experimental version of QR-Code compaction 

AlexGeller1:
Added experimental version of QR-Code compaction which can be activated via EncodeHintType.QR_COMPACT

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
+++ b/core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
@@ -49,6 +49,8 @@ public final class Encoder {
       25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, -1, -1, -1, -1, -1,  // 0x50-0x5f
   };
 
+//@Sean: Is the default encoding indeed ISO-8859-1? I also implemented the MinimalEncoder assuming this but now I looked at the specification and find
+//       in "8.3.1 Extended Channel Interpretation (ECI) Mode" the sentence "The default interpretation for QR Code is ECI 000020 representing the JIS8 and Shift JIS character sets".
   static final Charset DEFAULT_BYTE_MODE_ENCODING = StandardCharsets.ISO_8859_1;
 
   private Encoder() {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
```diff
@@ -49,6 +49,8 @@
       25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, -1, -1, -1, -1, -1,  // 0x50-0x5f
   };
 
+//@Sean: Is the default encoding indeed ISO-8859-1? I also implemented the MinimalEncoder assuming this but now I looked at the specification and find
```
srowen, line 52: That's what 6.4.5 in the 2006 spec says. Which version are you looking at? I don't see an 8.3.1 in my copy.
In any event I wouldn't change this behavior.

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,736 @@
+package com.google.zxing.qrcode.encoder;
```
srowen, line 1: Copyright header here

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,736 @@
...
+ *
+ * But as mentioned, it is not clear to me if this can ever happen with the actual values.
+ *
+ * ECI switching:
+ *
+ * In multi language content the algorithm selects the most compact representation using ECI modes. For example the
+ * it is more compactly represented using one ECI to UTF-8 rather than two ECIs to ISO-8859-6 and ISO-8859-1 if the text contains more ASCII characters (since they are represented as
+ * one byte sequence) as opposed to the case where there are proportionally more Arabic characters that require two bytes in UTF-8 and only one in ISO-8859-6.
+ */
+public class MinimalEncoder {
```
srowen, line 58: Nit: Can we keep this package-private? and final

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,736 @@
...
+
+    static final boolean PRINT_STATS = false; //print memoization memory usage
+
+    enum VersionSize {
+        Small,
```
srowen, line 63: Enums should be capitalized.

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,736 @@
...
+        Medium,
+        Large;
+        public String toString() {
+            return "Small".equals(name()) ? "version 1-9" : "Medium".equals(name()) ? "version 10-26" : "version 27-40";
```
srowen, line 67: Just make this a member of the enum like 'description' and return it in toString

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,736 @@
...
+        }
+    }
+
+    String stringToEncode;
```
srowen, line 71: Likewise private

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,736 @@
...
+//Average usage density should perhaps be monitored to see if not a hashtable is preferable. The GS1 test example uses 67% of an array of 1727 entries. Encoding the URL 'http://www.google.com'
+//uses 27% of an array of size 756.
+//the dimension sizes of the array are:
+//- position-in-input: length of input in characters
+//- character-encoding: minimally 3 (ISO-8859-1, UTF-8 and UTF-16). The input is scanned up front and for example ISO-8859-6 is added if an arabic character is encountered.
+//- version-class: exactly 3 representing version 1-9, 10-26 and 27-40.
+//- Mode: exactly 4 (representing the modes KANJI, ALPHANUMERIC, NUMERIC and BYTE)
+    ResultNode[][][][] memoizedResults; //[position-in-input][character-encoding][version-class][Mode]
+
+/**Encoding is optional (default ISO-8859-1) and version is optional (minimal version is computed if not specified*/
```
srowen, line 86: Fix indent

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,736 @@
...
+    MinimalEncoder(String stringToEncode,Version version,boolean isGS1) {
+        this.stringToEncode = stringToEncode;
```
srowen, line 88: 2 space indent

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,736 @@
...
+            if (isoEncoders[j] != null) {
+                encoders[index++] = isoEncoders[j];
+            }
+        }
+        encoders[index++] = StandardCharsets.UTF_8.newEncoder();
+        encoders[index++] = StandardCharsets.UTF_16BE.newEncoder();
+        
+        memoizedResults = new ResultNode[stringToEncode.length()][encoders.length][3][4];
+    }
+    public static void main(String[] args) throws Exception {
```
srowen, line 142: Space between methods
But we don't need a main() runner method here. The CLI is in another class

srowen (changes requested):
Just a few initial comments on code cleanup; I haven't looked in detail yet

core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
AlexGeller1: Good, mine is older. Is there a way that I can get your copy?

srowen: I don't believe copies of the ISO spec are supposed to be shared as they are, weirdly, for sale - if it comes up somehow I can provide some info from it

AlexGeller1: Thanks, that's very good. Something else, there seem to be different opinions on who presses "Resolve conversation", the request opener or the reviewer. The git hub documentation says the request opener so that I will now mark those things as resolved that I think are resolved. If you handle it differently, please tell me.

core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
```diff
@@ -78,64 +78,87 @@ public static QRCode encode(String content,
...
+
+    boolean hasGS1FormatHint = hints != null && hints.containsKey(EncodeHintType.GS1_FORMAT) && Boolean.parseBoolean(hints.get(EncodeHintType.GS1_FORMAT).toString());
+    boolean hasCompactionHint = hints != null && hints.containsKey(EncodeHintType.QR_COMPACT) && Boolean.parseBoolean(hints.get(EncodeHintType.QR_COMPACT).toString());
+
+    if (hasCompactionHint) {
+        mode = Mode.BYTE;
+
+        MinimalEncoder.ResultNode rn = MinimalEncoder.encode(content,null,hasGS1FormatHint);
+        while (!willFit(rn.getSize(), rn.getVersion(ecLevel), ecLevel)) {
+            if (rn.getVersion(ecLevel).getVersionNumber() <= 26) {
```
srowen, line 93: 2-space indent everywhere

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,778 @@
...
+
+/**Encoding is optional (default ISO-8859-1) and version is optional (minimal version is computed if not specified*/
+  MinimalEncoder(String stringToEncode,Version version,boolean isGS1) throws WriterException {
+    this.stringToEncode = stringToEncode;
+    if (version != null) {
+      this.version = version;
+    }
+    this.isGS1 = isGS1;
+    CharsetEncoder[] isoEncoders = new CharsetEncoder[15];
+    isoEncoders[0] = StandardCharsets.ISO_8859_1.newEncoder();
```
srowen, line 107: Use line break to break up blocks more

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,778 @@
...
+    
+
+/**Encoding is optional (default ISO-8859-1) and version is optional (minimal version is computed if not specified*/
+  MinimalEncoder(String stringToEncode,Version version,boolean isGS1) throws WriterException {
+    this.stringToEncode = stringToEncode;
+    if (version != null) {
+      this.version = version;
+    }
+    this.isGS1 = isGS1;
+    CharsetEncoder[] isoEncoders = new CharsetEncoder[15];
```
srowen, line 106: What's magical about 15 here?

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,778 @@
...
+  void addEdge(Vector<ResultNode>[][][] vertices,ResultNode rn,ResultNode previous) {
+    int vertexIndex = rn.position + (rn.mode == Mode.ECI ? 1 : getEncodingGranularity(rn.mode));
+    if (vertices[vertexIndex][rn.charsetEncoderIndex][getCompactedOrdinal(rn.mode)] == null) {
+      vertices[vertexIndex][rn.charsetEncoderIndex][getCompactedOrdinal(rn.mode)] = new Vector<ResultNode>();
+    }
+    vertices[vertexIndex][rn.charsetEncoderIndex][getCompactedOrdinal(rn.mode)].add(rn);
+    rn = rn.getHead();
+    rn.setPrevious(previous);
+
+    if (DEBUG) {
```
srowen, line 322: I generally have not put logging in the core library - I'd just remove this

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,778 @@
...
+              int toPosition = fromPosition + getEncodingGranularity(rn.mode);
+              System.err.println("DEBUG: (" + fromKEY + ") -- " + rn.mode + (toPosition - fromPosition > 0 ? "(" + stringToEncode.substring(fromPosition, toPosition) + ")" : "") + " (" + rn.getSize(true) + ")" + " --> " + "(" + vertexKey + ")");
+            }
+          }
+        }
+      }
+    }
+  }
+
+/** encodes minimally using Dijkstra.*/
```
srowen, line 396: These javadocs need to be indented to match what they refer to

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,778 @@
...
+//The lists are created lazily
+    Vector<ResultNode>[][][] vertices = new Vector[inputLength + 1][encoders.length][4];
+    addEdges(version,vertices,0,null);
+    if (DEBUG) {
+        System.err.println("DEBUG Initial situation");
+        printEdges(vertices);
+    }
+    for (int i = 1; i <= inputLength; i++) {
+      for (int j = 0; j < encoders.length; j++) {
+        for (int k = 0; k < 4; k++) {
```
srowen, line 497: Same, what's 4 here

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -0,0 +1,778 @@
...
+        throw new WriterException("Internal error: failed to encode");
+    }
+  }
+    
+  byte[] getBytesOfCharacter(int position,int charsetEncoderIndex) {
+    //TODO: Is there a more efficient way for a single character?
+    return stringToEncode.substring(position,position + 1).getBytes(encoders[charsetEncoderIndex].charset());
+ }
+
+  class ResultNode {
```
srowen, line 568: private final?

core/src/main/java/com/google/zxing/qrcode/encoder/QRCode.java
```diff
@@ -38,6 +38,9 @@ public QRCode() {
     maskPattern = -1;
   }
 
+  /**
+   * @return the mode. Not relevant if {@link com.google.zxing.EncodeHintType.java#QR_COMPACT} is selected.
```
srowen, line 42: Why isn't it relevant - it still selects some mode no?

core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
AlexGeller1: Arghh, sorry. A pity that the style tool doesn't check that too.

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
AlexGeller1: 15 is the maximum number of ISO-8859 encoders (1 through 16 and 12 doesn't exist but I am not compacting in the array)

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
AlexGeller1: There are only 4 modes (KANJI, ALPHANUMERIC, NUMERIC and BYTE) that are relevant in the algorithm. There is a function getCompactedOrdinal(Mode mode) that "compacts" the ordinals of these to values between 0 and 3.

core/src/main/java/com/google/zxing/qrcode/encoder/QRCode.java
AlexGeller1: Yes, it sets it to a constant value (currently Mode.BYTE). I could make it return the value of the first mode value in the bits from the left. My thinking was, that asking for "the" mode in a multi mode code doesn't make much sense so that added that comment.

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
AlexGeller1: An instance of this class is returned as a result of calling MinimalEncoder.encode(). It has methods to retrieve the bits and the version that was chosen so that it can't be private as long as we call it from Encoder. 
I now made it final and made all members and method private but for these (they are called from Encoder):
void getBits(BitArray bits)
Version getVersion(ErrorCorrectionLevel ecLevel)
int getSize()
toString has to be public and it is called from the unit test.

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
AlexGeller1: Are you sure? I find debugging the algorithm without it, difficult.

AlexGeller1: How about commenting it?

srowen: OK, comment it, or remove before the final merge

core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
srowen: I just added / enforced checks for line length and indent, FWIW. It may flag some of this PR now

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
srowen: OK can it be static?

srowen:
Sorry I just made a minor merge conflict for you here

core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
AlexGeller1: Not currently because it accesses non static methods from the enclosing class which in turn access the input string and the encoders (the method ResultNode.getBits() for example needs to access the input).

core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
AlexGeller1: I failed to figure out how I could update the branch from master so I changed pom.xml and checkstyle.xml temporarily the way you did and fixed/pushed the new issues on Encoder and MinimalEncoder that it found. I also fixed the indentation of Encoder.willFit() so that there shouldn't be a merge conflict later.

AlexGeller1:
I don't think that I am allowed to fix the reintegration merge conflict of the failing push 4130f24 (I explained above how it came about). At least I don't find an option that would allow me to do that.

srowen:
You just need to rebase/merge your branch on master and push it

AlexGeller1:
Sean, there is a big bug. I fixed it. I am just fixing the test case. One moment please, OK?

AlexGeller1:
I have to go. I will commit with commented out test cases and finish fixing them when I come back later, OK?

PR #1446, opened 2021-10-01
fix(core): do not evaluate out of index as white on qrcode mask evaluation

0xWOF:
## Background
By `ISO/IEC 18004:2015` standard, 40 points is added when `10111010000` or `00001011101` binary string is existed horizontally or vertically on qrcode matrix.
But on `qrcode/encoder/MaskUtil.java`'s `applyMaskPenaltyRule3` method add 40 points when `1011101` binary string is on edge of matrix.
For example, below matrix's points should be 40, but result of `applyMaskPenaltyRule3` method is 80.
```
Matrix is 11 x 1
10111010000
```
Cause of is problem is `applyMaskPenaltyRule3` method evaluate out of index as white.
I think this problem is created with this [commit](https://github.com/zxing/zxing/commit/558dad051e7bfc0940f6bdf5dff928edd30851b4).

## Change
Modify `isWhiteHorizontal` and `isWhiteVertical` method to do not evaluate out of index as white.
Add unit test for problem.
Fix unit test for qrcode encoder. (2 unit test is same with before this [commit](https://github.com/zxing/zxing/commit/558dad051e7bfc0940f6bdf5dff928edd30851b4))

## Check
I checked `core` test code is passed.

## Other
Even this problem, result of qrcode can be scanned. Because qrcode mask is not affected to data.
Therefore this pull request have potential risk but there are no benefit except it better follow standard.


srowen:
I see, I suspect you're right here. Do the resulting masks have a better score? that would be strong evidence this is a good fix. (Or maybe the point is, this fixes the scoring.) While i hesitate to change behavior that's been there for 8 years, I think this is probably right.

0xWOF:
> Do the resulting masks have a better score?

Result of masks are have lower score. Because the issue add 40 penalty score for incorrect case. 
Below data is score for each masks when run `QRCode qrCode = Encoder.encode("ABCDEF", ErrorCorrectionLevel.H);` code.
(same with (core.qrcode.encoder.EncoderTestCase testEncode case))
```
Before
0 1110 | 1 1064 | 2 1148 | 3 1139 | 4 1055 - Best | 5 1124 | 6 1139 | 7 1076
After
0 310 - Best | 1 464 | 2 548 | 3 379 | 4 495 | 5 564 | 6 499 | 7 396
```
Each score of masks are lower than before, and lowest mask is changed. (4 => 0)

> While i hesitate to change behavior that's been there for 8 years

I totally agree with your opinion, if you think this PR is too risky, please feel free to close PR.
Or if you think this PR have some merit to merge, what can i do for make concrete confidence? 

ex. add more test cases, or add some end to end tests (encode to decode), ...

srowen:
Do you happen to know any other reference implementation we might compare to? if not, no big deal, I tend to take your word on it if you have researched the issue. I have not!

0xWOF:
Oh, that great idea. Then I'll check some several other qrcode implementations and compare the result score of masks. 

But I worry that zxing port implementations almost have same issue and some other implementations are follow `ISO/IEC 18004:2000` standard (only check `1011101` binary string for score3). So I think there are few implementations to fit on compare.

I'll comment about this PR is safe to merge or not after I finish compare.

thanks!

srowen:
Apologies, there's a merge conflict now - just rebase/merge your branch.

If you can't find another reference impl, that's OK. I think this is probably correct

0xWOF:
Sorry for late reply.

I compare zxing after merge this PR with 2 repository <https://github.com/soldair/node-qrcode> and <https://github.com/Bacon/BaconQrCode>
Other repositories penalty3 score logic is same with zxing after merge this PR.
So I think this PR is safe to merge!

---

## node-qrcode

I test this repository with below code and result is below.
Current zxing choose maskPattern 4 and after merge PR choose maskPattern 0.
And this repository choose maskPattern 0.
Also I print Penalty3 score on zxing and this repository.
After zxing merge PR scores and this repository is same.

### code
```
const qrcode = require('qrcode')

console.log(
    qrcode.create(
        [{ data: 'ABCDEF', mode: 'alphanumeric' }],
        { errorCorrectionLevel: 'H' }
    )
)
```

### result
```
{
    ...
    version: 1,
    errorCorrectionLevel: { bit: 2 },
    maskPattern: 0,
    ...
}
```
### Penalty3 Scores for each masks
```
40 / 120 / 200 / 80 / 160 / 200 / 120 / 80
```


## BaconQrCode

I check this repository's mask scoring [code](https://github.com/Bacon/BaconQrCode/blob/f73543ac4e1def05f1a70bcd1525c8a157a1ad09/src/Encoder/MaskUtil.php#L98). And I see same `if` condition on this PR.
I do not test this repository. (Because I'm not familiar on PHP) 
But I think this will work same with after merge this PR.

PR #1449, opened 2021-10-10
Fix qr compaction

AlexGeller1:


AlexGeller1:
Hi Sean,
this is not related to this pull request but on the topic in general.
1) I saw that relatively recent in CharacterSetECI.java some encodings like ISO-8859-6 were commented out. Do I understand things right that Charset.forName() throws an exception for that charset on some platforms and therefore they were removed.
If that is the case then the QR-code implementation should encode the ECI code without the help of this class, no? Right now I am using this class to obtain the ECI value so that ISO-8859-6 and the other values are not supported.
2) Regarding changing the default value of QR_COMPACT from false to true there is perhaps a middle ground solution. One could scan the input string like it is done right now in Encoder.chooseMode() and iff all characters are from one class (BYTE, ALPHANUMERIC, NUMERIC or KANJI) then dispatch to the the fast single mode encodation and otherwise use the compacting version.

srowen:
Yes those character sets are not guaranteed to be supported by the JDK, and looks like they are not in practice. If so, then it's not possible to decode segments encoded that way. You could argue that decoding should still succeed and return no parsed string, just bytes, but, I think the caller almost never wants that, and these encodings are niche (JDKs don't bother to support it) so I think it's just 'unsupported'.

AlexGeller1:
Hi Sean,
are you waiting for action from my side?
If not ,no worries but just in case that it is about the LGTM warning about the useless comparisons, I had committed a fix to that very shortly before you pointed me to it.

PR #1450, opened 2021-10-13
Add support for EncodeTypeHint.CHARACTER_SET for compacted QR-Codes

AlexGeller1:


AlexGeller1:
I hope that you are OK with me reinstituting some of the asserts. They assert invariants and there is no code handling the case of failure (some of the debugging code was prepared to work around it but that code has been removed).

AlexGeller1:
The improvements are:
- The code is now about 60 lines shorter
- The algorithm function "encodeSpecificVersion()" now uses a class called Edge that has no methods and returns the solution in form of a list of these.
- The ResultList now contains exactly one ResultNode per mode in the solution. This was not the case in the previous version.

PR #1451, opened 2021-10-15
Change sort method to work with any android api

caovanthanh203:
List.sort(Comparator) only work with android API >= N
So I change it to Collections.sort(List, Comparator)


srowen:
There is more that would have to change besides this. This library requires Java 8, which is itself even EOL. So you do need semi-recent versions of Android to use this code

PR #1452, opened 2021-10-18
Compacted QR-code: remove needless adding of Mode.TERMINATOR (is taken care of in Encoder.terminateBits)

AlexGeller1:


AlexGeller1:
What are your thoughts regarding the non-compacted Qr-Codes? As far as I can tell, they never include a terminator.
It wouldn't be a big deal to make them have the same feature and share the "capacityInBits()" function.
I see the following pros:
- Both version are spec conform.
- For a single mode symbol (e.g. all BYTE or all NUMERIC) the output of the two encoders will be identical.
- Right now a single mode symbol is non-compacted smaller than it's compacted counterpart by 4 bit.
And the following cons: 
- The output will change so that for any input and the codes will be larger by 4 bit.

AlexGeller1:
My bad. The terminator is indeed added in Encoder.terminateBits. I added a comment so that a grep for "TERMINATOR" will now find the place where it is done.

Issue #624, opened 2016-06-30
Illegal Character Exception


bomski:
Hi everyone,

i always get a illegal character exception if i try to create a DataMatrix with following code

```
HashMap<EncodeHintType, Object> options = new HashMap<>();
options.put(EncodeHintType.DATA_MATRIX_SHAPE, SymbolShapeHint.FORCE_SQUARE);

int width = 128;
int height = 128;
String content = "test TE>240 2 I.E ST>300";

BitMatrix bitMatrix = new DataMatrixWriter().encode(content, BarcodeFormat.DATA_MATRIX, width, height, options)
```

**Stacktrace**

```
java.lang.IllegalArgumentException: Illegal character: . (0x002e)
    at com.google.zxing.datamatrix.encoder.HighLevelEncoder.illegalCharacter(HighLevelEncoder.java:445)
    at com.google.zxing.datamatrix.encoder.X12Encoder.encodeChar(X12Encoder.java:65)
    at com.google.zxing.datamatrix.encoder.X12Encoder.encode(X12Encoder.java:34)
    at com.google.zxing.datamatrix.encoder.HighLevelEncoder.encodeHighLevel(HighLevelEncoder.java:182)
    at com.google.zxing.datamatrix.DataMatrixWriter.encode(DataMatrixWriter.java:84)
    at com.google.zxing.Writer$encode.call(Unknown Source)
```

Is this a bug or i'm missing something?

tanelihuuskonen:
As far as I can tell, it's a bug.

srowen:
Agree, it looks like a problem with `HighLevelEncoder.lookAheadTest`. It switches to X12 encodation when it can't encode ".". I debugged it for a bit but am not familiar with this part of the code. It does look like the part of the code labelled "Step R" does not follow the Data Matrix spec, quite. I hacked a quick fix for that, but it caused other test failures.

```
        if (minCount == 1 && mins[EDIFACT_ENCODATION] > 0) {
          return EDIFACT_ENCODATION;
        }
        if (minCount == 1 && mins[TEXT_ENCODATION] > 0) {
          return TEXT_ENCODATION;
        }
        if (minCount == 1 && mins[X12_ENCODATION] > 0) {
          return X12_ENCODATION;
        }
```
should be
```
        if (intCharCounts[EDIFACT_ENCODATION] + 1 < intCharCounts[BASE256_ENCODATION]
[...]

PR #1454, opened 2021-10-29
Fix bug624: Illegal Character Exception

AlexGeller1:


AlexGeller1:
Originally I had doubts that the comparisons have to be done with rounded integers because the specification doesn't say so. I don't have those doubts any more because I wrote a small program that shows for three example patterns that the lookahead algorithm proposes incorrect encoding (an encoding in which the character heading the prefix cannot be encoded) if you use float point comparisons. It also shows that you can exit B256 mode only if you have a digit in the prefix. This is bad because it means that it will for example never choose EDIFACT encoding now matter how long the data is that follows the part that needs to be B256 encoded. I therefore think that the code was fine but for the fixes that I made in this PR.

core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
```diff
@@ -295,26 +295,40 @@ static int lookAheadTest(CharSequence msg, int startpos, int currentMode) {
...
         byte[] mins = new byte[6];
         findMinimums(charCounts, intCharCounts, Integer.MAX_VALUE, mins);
-        int minCount = getMinimumCount(mins);
 
-        if (intCharCounts[ASCII_ENCODATION] < intCharCounts[BASE256_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[C40_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[TEXT_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[X12_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[EDIFACT_ENCODATION]) {
+        if (intCharCounts[ASCII_ENCODATION] + 1 <= intCharCounts[BASE256_ENCODATION]
```
srowen, line 299: Here, "+ 1 <=" is the same as "<" so we can leave it.
Below, is "+ 1 <" intended over <= ? just checking. It's different in one line.
Also I wonder if we could wrap this up in a utility method that just finds the minimum count, but maybe the point is that the logic is slightly different each time

pzygielo:
Just to ease navigation:
- #624

core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
```diff
@@ -295,32 +295,34 @@ static int lookAheadTest(CharSequence msg, int startpos, int currentMode) {
...
         byte[] mins = new byte[6];
         findMinimums(charCounts, intCharCounts, Integer.MAX_VALUE, mins);
-        int minCount = getMinimumCount(mins);
 
-        if (intCharCounts[ASCII_ENCODATION] < intCharCounts[BASE256_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[C40_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[TEXT_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[X12_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[EDIFACT_ENCODATION]) {
+        if (intCharCounts[ASCII_ENCODATION] < min(intCharCounts[BASE256_ENCODATION]
```
srowen, line 299: So there's no "+ 1" with "<" in this case, unlike the others?
It feels like we could refactor this further - given an index in intCharCounts, say whether it's more than 1 smaller than other indices?

AlexGeller1: Well in the spec all is written as a + 1 < b or a + 1 <= b. But as you pointed out, a + 1 <= b can be rewritten as a < b so I changed that. 
The situation is as follows:
We have one case where we compare +1 <= against all others.
Then we have a second case where we compare <= against one or +1 < against 4 other values
Then we have three cases where we compare + 1 < against 5 other values and finally a case where we compare + 1 < against 4 other values.
Since sometimes you have to compare against 4 and sometimes against 5 values, I don't see how the signature of a function that works with indexes into intCharCounts could look good and understandable. Maybe I am overlooking something. Can you be more concrete?

srowen: OK I get it, so "+ 1 <" is intended, just checking.
OK if the logic varies that much, this may be as good as it gets. The min() function looks a little clunky and could be written as varargs instead... but may be a little slower, so, no big deal

core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
```diff
@@ -295,32 +295,34 @@ static int lookAheadTest(CharSequence msg, int startpos, int currentMode) {
...
+            , intCharCounts[C40_ENCODATION], intCharCounts[TEXT_ENCODATION], intCharCounts[X12_ENCODATION]
```
srowen, line 300: commas on previous line

I tend to write

```
if (foo ||
    (bar &&
     baz)) {
```

core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
```diff
@@ -342,6 +344,12 @@ static int lookAheadTest(CharSequence msg, int startpos, int currentMode) {
       }
     }
   }
+  private static int min(int f1, int f2, int f3, int f4, int f5) {
+    return Math.min(min(f1, f2, f3, f4),f5);
+  }
+  private static int min(int f1, int f2, int f3, int f4) {
```
srowen, line 350: (Blank line in between)

core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
AlexGeller1: So do I, but the original code was like that. I will change it.

srowen: You can leave or change whether the operator is on a new line - whatever looks better to you and is consistent

AlexGeller1:
Hi Sean,
I haven't head from you in a while.
Do you have doubts regarding the fix?
FYI, I have spotted a problem in the specification that leads to a very similar error and it turns out that ZXing also suffers from this problem with slightly different patterns that lead to failure.
I contacted people from ISO and it turns out that zint also suffers from this problem. Martin Burke proposed a fix that I could also put in place in ZXing (See: https://sourceforge.net/p/zint/code/ci/68566fefd2a4623041b0232a5941f8fed1e22018/)

core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
```diff
@@ -295,32 +295,34 @@ static int lookAheadTest(CharSequence msg, int startpos, int currentMode) {
...
         findMinimums(charCounts, intCharCounts, Integer.MAX_VALUE, mins);
-        int minCount = getMinimumCount(mins);
 
-        if (intCharCounts[ASCII_ENCODATION] < intCharCounts[BASE256_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[C40_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[TEXT_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[X12_ENCODATION]
-            && intCharCounts[ASCII_ENCODATION] < intCharCounts[EDIFACT_ENCODATION]) {
+        if (intCharCounts[ASCII_ENCODATION] < min(intCharCounts[BASE256_ENCODATION],
+            intCharCounts[C40_ENCODATION], intCharCounts[TEXT_ENCODATION], intCharCounts[X12_ENCODATION],
```
srowen, line 300: I think we need a continuation indent here, one deeper, unless the style checker doesn't like it. Same below

srowen (commented):
Looks Ok to me, just see if we can improve the indent of the nested method call

PR #1465, opened 2021-12-08
Simplified code, improved memory usage and performance of the QR-Code minimal encoder.

AlexGeller1:


core/src/main/java/com/google/zxing/qrcode/encoder/MinimalEncoder.java
```diff
@@ -304,15 +304,16 @@ static int getCompactedOrdinal(Mode mode) {
     }
   }
 
-  void addEdge(List<Edge>[][][] edges, int position, Edge edge) {
+  void addEdge(Edge[][][] edges, int position, Edge edge) {
     int vertexIndex = position + edge.characterLength;
-    if (edges[vertexIndex][edge.charsetEncoderIndex][getCompactedOrdinal(edge.mode)] == null) {
-      edges[vertexIndex][edge.charsetEncoderIndex][getCompactedOrdinal(edge.mode)] = new ArrayList<>();
+    if (edges[vertexIndex][edge.charsetEncoderIndex][getCompactedOrdinal(edge.mode)] == null ||
```
srowen, line 309: You can save off the result of dereferencing the first two dimensions, and reuse it 3 places here. You can probably save off more intermediate vars to simplify too

AlexGeller1: Thanks for the quick validation. FYI, I found this obvious simplification while writing the minimal encoder for data matrix and was faced with the problem of implementing it efficiently in C. If you are interested, then I can make it available for XZing. I don't know it you are aware, but the minimal encoding algorithm in the data matrix specification (Annex P) is not as good as the QR-code one. For example, when in mode B256, the algorithm will exit this mode only if it encounters two consecutive digits. This means that something that starts with some umlauts and then is followed by uppercase characters will be encoded with 8 bits per character while all the uppercase characters could be encoded per C40 or X12 with 5.33 bits per character.

srowen (approved):

PR #1479, opened 2022-01-11
Minimal encoding for Data Matrix with multi ECI and GS1-FNC1 support (2nd)

AlexGeller1:
Feature is off by default and can be activated by EncodeTypeHint.DATA_MATRIX_COMPACT

### DIFF
--- a/core/src/main/java/com/google/zxing/EncodeHintType.java
+++ b/core/src/main/java/com/google/zxing/EncodeHintType.java
@@ -46,10 +46,15 @@ public enum EncodeHintType {
 
   /**
    * Specifies whether to use compact mode for Data Matrix (type {@link Boolean}, or "true" or "false"
-   * Please note that when compaction is performed, the most compact character encoding is chosen
-   * for characters in the input that are not in the ISO-8859-1 character set. Based on experience,
-   * some scanners do not support encodings like cp-1256 (Arabic). In such cases the encoding can
-   * be forced to UTF-8 by means of the {@link #CHARACTER_SET} encoding hint.
+   * The compact encoding mode also supports the encoding of characters that are not in the ISO-8859-1
+   * character set via ECIs.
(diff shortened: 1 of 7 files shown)


srowen (commented):
Looks OK if tests pass. If there are minor tweaks to style I'd recommend I can sort them out later.

AlexGeller1:
There are three @Seans in the code where I had doubts and wanted your opinion. There are two in HighLevelEncodeTestCase and one in DataMatrixWriter. That one is important since I need to know how users encode FNC1 in input strings. I implemented it by the GS (ASCII 29) character but perhaps it is done differently by some other escape mechanism in ZXing.

AlexGeller1:
One other thing, MACRO5 and MACRO6 isn't implemented yet. It strips the '[)>' stuff but doesn't insert the values. That needs to be done still.

srowen:
I don't know the details of that enough to comment - have forgotten or never knew it. I'd say in general, leave existing behavior unless sure it should be something different; implement functionality that is coherent in itself even if some parts could be extended, rather than leave something partly implemented but not quite working. Up to your judgment from there.

AlexGeller1:
I looked at the source Code128Writer. Apparently it uses the character 0xf1 to encode FNC1 in input. What I have implemented is more what is in place for QR-Code (there the AI it is the GS character because GS1 says to use it instead of FNC1). For any bat code GS says that the decoder needs to yield GS for FNC1 on decoding so that using it for encoding makes sense in my eyes. The code-128 0xf1 seems odd to me but perhaps it was used because code-128 additionally has FNC2, FNC3 and FNC4 which are encoded by 0xf2 through 0xf4.

AlexGeller1:
> Looks OK if tests pass. If there are minor tweaks to style I'd recommend I can sort them out later.

Yes, all test pass.

srowen (approved):

PR #1483, opened 2022-01-14
Minimal encoding for Code-128

AlexGeller1:
Added CODE128_COMPACT encoding hint for activation.

### DIFF
--- a/core/src/main/java/com/google/zxing/oned/Code128Writer.java
+++ b/core/src/main/java/com/google/zxing/oned/Code128Writer.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Arrays;
 
 /**
  * This object renders a CODE128 code as a {@link BitMatrix}.
@@ -39,6 +40,7 @@ public final class Code128Writer extends OneDimensionalCodeWriter {
   private static final int CODE_CODE_B = 100;
   private static final int CODE_CODE_C = 99;
   private static final int CODE_STOP = 106;
+  private static final int CODE_SHIFT = 98;
 
   // Dummy characters used to specify control characters in input
   private static final char ESCAPE_FNC_1 = '\u00f1';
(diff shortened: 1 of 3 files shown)


AlexGeller1:
I just saw that the diff for the method Code128Writer.chooseCode() looks huge but I didn't change it at all (0 lines). I just moved it into a Selector class so that I can choose between this implementation and the optimal implementation.

core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -179,26 +191,26 @@
...
-              default:
-                // CODE_CODE_C
-                if (position + 1 == length) {
-                  // this is the last character, but the encoding is C, which always encodes two characers
-                  throw new IllegalArgumentException("Bad number of characters for digit only encoding.");
-                }
-                patternIndex = Integer.parseInt(contents.substring(position, position + 2));
-                position++; // Also incremented below
-                break;
+            if (codeSet == CODE_CODE_A ||
```
srowen, line 194: Why remove the switch?

core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -282,66 +302,239 @@ private static CType findCType(CharSequence value, int start) {
...
-        if (c < ' ' || (oldCode == CODE_CODE_A && (c < '`' || (c >= ESCAPE_FNC_1 && c <= ESCAPE_FNC_4)))) {
-          // can continue in code A, encodes ASCII 0 to 95 or FNC1 to FNC4
-          return CODE_CODE_A;
+  }
+
+  /** 
+   * Encodes minimally using Divide-And-Conquer with Memoization
+   **/
+  private static class MinimalSelector implements Selector {
+    enum Charset { A, B, C, NONE };
```
srowen, line 377: private?

core/src/main/java/com/google/zxing/oned/Code128Writer.java
AlexGeller1: Because the cases are are no longer only an integer

core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -179,26 +185,28 @@
...
-              default:
-                // CODE_CODE_C
-                if (position + 1 == length) {
-                  // this is the last character, but the encoding is C, which always encodes two characers
-                  throw new IllegalArgumentException("Bad number of characters for digit only encoding.");
-                }
-                patternIndex = Integer.parseInt(contents.substring(position, position + 2));
-                position++; // Also incremented below
-                break;
+            if ((codeSet == CODE_CODE_A &&
```
srowen, line 188: OK I see. I'm a little concerned about changing the exisitng code path; adding a second one isn't as big a deal. Are these changes necessary vs just having an alternative impl?

core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -282,66 +298,242 @@ private static CType findCType(CharSequence value, int start) {
...
-        char c = value.charAt(start);
-        if (c < ' ' || (oldCode == CODE_CODE_A && (c < '`' || (c >= ESCAPE_FNC_1 && c <= ESCAPE_FNC_4)))) {
-          // can continue in code A, encodes ASCII 0 to 95 or FNC1 to FNC4
-          return CODE_CODE_A;
+  }
+
+  /** 
+   * Encodes minimally using Divide-And-Conquer with Memoization
+   **/
+  private static class MinimalSelector implements Selector {
```
srowen, line 372: Like I'd kind of prefer the separate new code live in its own top level class, for clarity

AlexGeller1: Sure, no problem. This version was the option with the most code reuse in my opinion. By reimplementing only the lookahead function it shared the checks (checks for maximum length and check for illegal characters), the the building of the pattern, the weighted checksum accumulation, and the adding of the checksum and the stop pattern to the patterns.
To make it less invasive I see two alternative options.
1) Initially I wrote a version (also without introducing a new top level class) that has two code paths that compute the "pattern" ArrayList. That way it also shared the checks and the adding of the stop pattern (the checksum adding could also be shared but then the function would need to return two values (the ArrayList and the checksum) .
2) We rename Code128Writer.java to Code128FastWriter.java and add a new class Code128MinimalWriter.java. The checking code could be put in a separate function in Code128FastWriter.java and then be reused by calling it from Code128MinimalWriter too.
I think that both options are fine. The checksum computation and the adding of the stop pattern are so trivial that I don't think that it is worth the bother to try to share them so that I lean slightly more towards option 2. 
Which would you prefer or do you see yet another good way of doing it?

AlexGeller1: Update: On second thought, I am not so fond of option 2 after looking at the code some more. I propose to come up with a solution that leaves the exiting encode() function completely without modification and just dispatches between two versions of the function. This way the two functions can share the constants and the standard overrides of OneDimensionalCodeWriter. I will make a proposal and then we can discuss it from there.

AlexGeller1: Update: I created an new PR #1484 with a less intrusive version.

PR #1484, opened 2022-01-16
Minimal encoding for Code-128 (2nd, less intrusive version)

AlexGeller1:
Added CODE128_COMPACT encoding hint for activation.

### DIFF
--- a/core/src/main/java/com/google/zxing/oned/Code128Writer.java
+++ b/core/src/main/java/com/google/zxing/oned/Code128Writer.java
@@ -72,6 +72,16 @@ public final class Code128Writer extends OneDimensionalCodeWriter {
 
   @Override
   protected boolean[] encode(String contents, Map<EncodeHintType,?> hints) {
+
+    int forcedCodeSet = check(contents, hints);
+
+    boolean hasCompactionHint = hints != null && hints.containsKey(EncodeHintType.CODE128_COMPACT) &&
+        Boolean.parseBoolean(hints.get(EncodeHintType.CODE128_COMPACT).toString());
+
+    return hasCompactionHint ? new MinimalEncoder().encode(contents) : encodeFast(contents, hints, forcedCodeSet);
+  }
+
+  protected static int check(String contents, Map<EncodeHintType,?> hints) {
     int length = contents.length();
     // Check length
     if (length < 1 || length > 80) {
@@ -139,6 +149,11 @@ public final class Code128Writer extends OneDimensionalCodeWriter {
           break;
       }
     }
+    return forcedCodeSet;
+  }
+
+  protected static boolean[] encodeFast(String contents, Map<EncodeHintType,?> hints, int forcedCodeSet) {
+    int length = contents.length();
 
     Collection<int[]> patterns = new ArrayList<>(); // temporary storage for patterns
     int checkSum = 0;
@@ -234,7 +249,10 @@ public final class Code128Writer extends OneDimensionalCodeWriter {
         checkWeight++;
       }
     }
+    return produceResult(patterns, checkSum);
+  }
 
+  static boolean[] produceResult(Collection<int[]> patterns, int checkSum) {
     // Compute and append checksum
     checkSum %= 103;
     patterns.add(Code128Reader.CODE_PATTERNS[checkSum]);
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -344,4 +362,205 @@ private static int chooseCode(CharSequence value, int start, int oldCode) {
     return CODE_CODE_B;
   }
 
+  /** 
+   * Encodes minimally using Divide-And-Conquer with Memoization
+   **/
+  private static class MinimalEncoder {
```
srowen, line 368: Could this be a top-level package private class? no big deal. whatever the other impl does.

core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -72,6 +72,16 @@
...
+
+    int forcedCodeSet = check(contents, hints);
+
+    boolean hasCompactionHint = hints != null && hints.containsKey(EncodeHintType.CODE128_COMPACT) &&
+        Boolean.parseBoolean(hints.get(EncodeHintType.CODE128_COMPACT).toString());
+
+    return hasCompactionHint ? new MinimalEncoder().encode(contents) : encodeFast(contents, hints, forcedCodeSet);
+  }
+
+  protected static int check(String contents, Map<EncodeHintType,?> hints) {
```
srowen, line 84: Can this be private?

core/src/main/java/com/google/zxing/oned/Code128Writer.java
AlexGeller1: Then I have to duplicate all the constants. As an inner class it shares them. I also have to remove the private from the methods check() and produceResult().

srowen: Oh OK that's fine. They can be pkg-private too, but that's reasonable. private inner classes can't see private methods?

AlexGeller1: Private inner classes can see them (they are private after the latest commit) but if I make it a top level class then I would need to make them pkg-private.

srowen: Right so how about the check() method

AlexGeller1: check() and produceResult() are both private after the latest commit,

AlexGeller1: >Oh OK that's fine. 

Does it mean that you are OK with duplicating the constants or does it mean that you are OK with leaving it as an inner class? I personally make any class that can be made as an inner class, an inner class, reasoning that it encapsulates better. I argue that if a class is only used from one class, why should any other class see it? 

Apart from that, the class MinimalEncoder is special in the sense that it could have been written completely as a single static function since the two members memoizedCost and minPath could have been arguments to that function. The only reason that I didn't do that was because I didn't want two more arguments on the stack since the function is recursive but that's just an optimization. The whole class could have been written as a single static function encodeMinimally() the same way that the existing encodeFast() function is written.

Something else, if we make it a top level class then we should also rename it in my opinion to something like Code128MinimalEncoder because the directory contains a number of different encoders (Code-39, Code-93, UPC/EAN and more) and the name "MinimalEncoder" wouldn't in my opinion make it clear enough what exactly it encodes minimally, agree?

srowen: Sorry I mean leave it as an inner class. I don't necessarily agree with making everything inner that can be but it's situational. static inner classes don't need to be inner and sometimes are such just for namespacing, but that's what packages are for IMHO. Sometimes they are intimately tied to the class so make sense. Scala has nicer finer-grained control that let top-level classes be more selectively private, but hey

Go for code simplicity and reuse while keeping the existing impl's behavior as as-is as possible

AlexGeller1: Thanks for clarifying. Is there anything else you want me to to on this one then? 
For summary, as I mentioned earlier the fast algorithm doesn't make uses of the SHIFT codeword so that big deviations in size can be obtained compared to the minimal encoder (shown in test Code128WriterTestCase.testShift() where the fast encoder produces a width of 1090 bars and the minimal encoder needs 253 less) . The relevance is however not clear to me since I don't know how much characters like newline are used in the field.
I however found that one of the test cases has a relevant difference. The test case Code128WriterTestCase.testRoundtrip() tests a typical GS1 bar code and the fast algorithm produces a slightly wider bar code than the minimal one. The difference is only 11 bars (total width is 166) and it may be the only case but it is there nevertheless.

AlexGeller1: Thanks for the clarification. Is there anything else you want me to do on this?
For summary, as I pointed out earlier the existing fast implementation does not make use of the SHIFT codeword so that large differences can be produced as shown in test case Code128WriterTestCase.testShift() where the minimal encoder encodes the same text in about 80% of the space. I however don't know how relevant that it as I don't know how much people in the field encode special characters like newline.
There is however one existing test case, namely the test Code128WriterTestCase.testRoundtrip() that is relevant because it is a standard GS1 example (digits and FNC1). The difference isn't big (11 bars) and perhaps it is the only case in which it happens but it is there nevertheless.

PR #1495, opened 2022-01-25
Added option to force C40 encoding in data matrix

AlexGeller1:
Forces C40 encoding in data-matrix via EncodeHintType.FORCE_C40 (apparently a French government requirement).
Note that not all input strings can be encoded in C40.

### DIFF
--- a/core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
+++ b/core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
@@ -125,7 +125,7 @@ public final class HighLevelEncoder {
    * @return the encoded message (the char values range from 0 to 255)
    */
   public static String encodeHighLevel(String msg) {
-    return encodeHighLevel(msg, SymbolShapeHint.FORCE_NONE, null, null);
+    return encodeHighLevel(msg, SymbolShapeHint.FORCE_NONE, null, null, false);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
```diff
@@ -137,15 +137,18 @@ public static String encodeHighLevel(String msg) {
    *                {@code SymbolShapeHint.FORCE_SQUARE} or {@code SymbolShapeHint.FORCE_RECTANGLE}.
    * @param minSize the minimum symbol size constraint or null for no constraint
    * @param maxSize the maximum symbol size constraint or null for no constraint
+   * @param forceC40 enforce C40 encoding
    * @return the encoded message (the char values range from 0 to 255)
    */
   public static String encodeHighLevel(String msg,
```
srowen, line 143: Let's keep an overload with the original signature, to avoid breaking the public API

core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
```diff
@@ -164,6 +167,13 @@ public static String encodeHighLevel(String msg,
     }
 
     int encodingMode = ASCII_ENCODATION; //Default mode
+
+    if (forceC40) {
```
srowen, line 171: Maybe put this init at the top, or even make a C40Encoder overload that carries the 'force' info so it can be used internally. I'd rather hide, I think, this detail inside the encoder rather than make callers do the work if possible. Like can C40Encoder call encodeMaximal internally?

and is forcing C40 really kinda synonymous with 'maximal' encoding?

srowen (changes requested):

core/src/main/java/com/google/zxing/datamatrix/encoder/HighLevelEncoder.java
AlexGeller1: >Like can C40Encoder call encodeMaximal internally

I don't see how but perhaps you have an idea. The problem is that we have to start in C40 encoding. If we act from inside C40Encoder.encode(), then we will get called when the lookahead test thinks that C40 encoding is appropriate. I was quite happy with the solution since it is very clear so see what happens if forceC40 is false (the important path). There is only one if that prunes the entire right at the start. 

>and is forcing C40 really kinda synonymous with 'maximal' encoding?
No, but the C40Encoder doesn't force c40 encoding. Instead it encodes maximally in the sense that it is greedy and encodes as much as it can when this function is called. In contrast, the normal encode() function just encodes on triplet and then consults the lookahead test to see if it should encode more.

srowen: OK I think it's fine if this is the cleanest solution

AlexGeller1:
The requester stackhchad reports that he has tested the PR and found it working as expected. If you don't have any objections then from my side this can be merged.

Issue #1502, opened 2022-02-21
Failure to scan some PDF417 codes


AlexGeller1:
This PDF417 bar code and some others fails to scan:
![image](https://user-images.githubusercontent.com/87009702/154929828-a030e6c8-618e-4851-909b-30d552b8c4b1.png)

PR #1503, opened 2022-02-25
Fix bug #1502 that PDF417 barcodes without initial latch to Text compaction mode fail to decode

AlexGeller1:
Text compaction mode is the default mode and does not need to be specified (See "5.4.2.1 Text Compaction sub modes"; quote: "The default compaction mode for PDF417 in effect at the start of each symbol shall always be Text compaction mode Alpha sub mode [uppercase alphabetic]")

### DIFF
--- a/core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
+++ b/core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
@@ -154,7 +154,6 @@ final class DecodedBitStreamParser {
           break;
       }
     }
-//@Sean: I propose to remove this. Why should an empty barcode be an error? This makes an empty string test fail.
     if (result.length() == 0 && resultMetadata.getFileId() == null) {
       throw FormatException.getFormatInstance();
     }
--- a/core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
+++ b/core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
@@ -104,11 +104,10 @@ final class DecodedBitStreamParser {
   static DecoderResult decode(int[] codewords, String ecLevel) throws FormatException {
     StringBuilder result = new StringBuilder(codewords.length * 2);
     Charset encoding = StandardCharsets.ISO_8859_1;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
```diff
@@ -154,12 +153,8 @@ static DecoderResult decode(int[] codewords, String ecLevel) throws FormatExcept
           codeIndex = textCompaction(codewords, codeIndex, result);
           break;
       }
-      if (codeIndex < codewords.length) {
-        code = codewords[codeIndex++];
-      } else {
-        throw FormatException.getFormatInstance();
-      }
     }
+//@Sean: I propose to remove this. Why should an empty barcode be an error? This makes an empty string test fail.
```
srowen, line 157: You can write these as comments in the PR, rather than code - you can remove this.
Is an empty barcode valid? it doesn't seem like it would ever be the correct result, even if technically allowed.
My concern was probably avoiding false positive scans, as it might be relatively easy to see a simple phantom barcode encoding the empty string.

AlexGeller1: >Is an empty barcode valid?

I read only the parts of the spec that deal with high level encoding and there I didn't come across anything that would forbid it. I can't however rule out that there is something in the spec in other places that forbids it.

>it doesn't seem like it would ever be the correct result, even if technically allowed

Agree. It just makes a test of mine fail. The background is, that I wrote a massive test that generates and encodes thousands of patterns and then compares the decoded result to the generated input for equivalence.
I think that it might also be a good idea to add this test to ZXing. The problem with the test is that is requires that the unit test can access the static methods pdf417.decoder.DecodedBitStreamParser.decode() and pdf417.encoder.PDF417HighLevelEncoder.encodeHighLevel(). Both classes and methods are declared as private so that that is not possible. Perhaps we can find a way to give the test access but prevent public access at the same time.

srowen: Let's keep it simple and as-is here, and not allow empty barcodes

PR #1504, opened 2022-02-28
Adding more PDF417 decoding tests

AlexGeller1:
The new tests generate codeword data with the high level encoder and check the size. The data is then used as input to test the high level decoder by verifying that the decoded result is identical with the input.
The test contains hard coded patterns, pattern generated by permuting characters of different character classes and a random binary test.

### DIFF
--- a/core/src/main/java/com/google/zxing/pdf417/encoder/PDF417HighLevelEncoderTestAdapter.java
+++ b/core/src/main/java/com/google/zxing/pdf417/encoder/PDF417HighLevelEncoderTestAdapter.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2022 ZXing authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.zxing.pdf417.encoder;
+
+import com.google.zxing.WriterException;
+
+import java.nio.charset.Charset;
+
+/**
+ * Test adapter for PDF417HighLevelEncoder to be called solely from unit tests.
+ */
+
+public final class PDF417HighLevelEncoderTestAdapter {
+
+  private PDF417HighLevelEncoderTestAdapter() {
+  }
+
+  public static String encodeHighLevel(String msg, Compaction compaction, Charset encoding) throws WriterException {
+    return PDF417HighLevelEncoder.encodeHighLevel(msg, compaction, encoding);
+  }
+}
--- a/core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
+++ b/core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
@@ -17,10 +17,16 @@
 package com.google.zxing.pdf417.decoder;
 
 import com.google.zxing.FormatException;
+import com.google.zxing.WriterException;
 import com.google.zxing.pdf417.PDF417ResultMetadata;
 import com.google.zxing.common.DecoderResult;
+import com.google.zxing.pdf417.encoder.Compaction;
+import com.google.zxing.pdf417.encoder.PDF417HighLevelEncoderTestAdapter;
+
 import org.junit.Assert;
 import org.junit.Test;
+import java.nio.charset.StandardCharsets;
+import java.util.Random;
 
 /**
  * Tests {@link DecodedBitStreamParser}.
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/google/zxing/pdf417/encoder/PDF417HighLevelEncoderTestAdapter.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Test adapter for PDF417HighLevelEncoder to be called solely from unit tests.
+ */
+
+public final class PDF417HighLevelEncoderTestAdapter {
+
+  private PDF417HighLevelEncoderTestAdapter() {
+  }
+
+  public static String encodeHighLevel(String msg, Compaction compaction, Charset encoding) throws WriterException {
```
srowen, line 32: For now can this just be a utility method in your new test class?

AlexGeller1: I am not sure that I understand the question. What are you referring to with "new test class"? 
The problem is that the unit test com.google.zxing.pdf417.decoder.PDF417DecoderTestCase cannot access the method com.google.zxing.pdf417.encoder.PDF417HighLevelEncoder.encodeHighLevel() because neither the class nor the method are public. I don't think that we should change that since the method is internal and not part of the public API. I therefore introduced the class com.google.zxing.pdf417.encoder.PDF417HighLevelEncoderTestAdapter that is a public proxy to that method. It contains a clear comment that hopefully keeps people from calling this method from their code. I could perhaps also add some code in the function that checks the call stack at runtime and refuses to to operate if it is called from anywhere else than from the unit test. The issue with that may be that it slows down the unit test since the called thousands of times from the unit test. I don't find this solution ideal but I don't have a better idea either. I am open to any other idea.

srowen: This new class you added right here.
Maybe I'm not reading this right but all classes in question are in the same package, so package-private access should be OK to call.

AlexGeller1: The test itself is in the package com.google.zxing.pdf417.decoder and it needs to call a method from com.google.zxing.pdf417.encoder.

AlexGeller1: Exactly, that's a good summary of the problem.

core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
```diff
@@ -175,4 +181,197 @@ public void testSampleWithNoDataNoMacro() throws FormatException {
...
+      codewords[i] = s.charAt(i - 1);
+    }
+    DecoderResult result = DecodedBitStreamParser.decode(codewords, "0");
+
+    assertEquals(input, result.getText());
+    return codewords.length;
+  }
+
+  private static int getEndIndex(int length, char[] chars) {
+    double decimalLength = Math.log(chars.length) / Math.log(10);
```
srowen, line 345: You can use Math.log10 for base-10 log

core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
```diff
@@ -175,4 +181,197 @@ public void testSampleWithNoDataNoMacro() throws FormatException {
...
+  }
+
+  private static void performPermutationTest(char[] chars,int length,int expectedTotal) throws WriterException,
+      FormatException {
+    int endIndex = getEndIndex(length, chars);
+    int total = 0;
+    for (int i = 0; i < endIndex; i++) {
+      total += encodeDecode(generatePermutation(i, length, chars));
+    }
+    String s = "['" + chars[0] + "'";
```
srowen, line 369: If this is at all perf-sensitive, it can use StringBuilders instead, but may not matter for a test

core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
```diff
@@ -175,4 +181,197 @@ public void testSampleWithNoDataNoMacro() throws FormatException {
...
+  @Test
+  public void testBinaryData() throws WriterException, FormatException {
+    byte[] bytes = new byte[500];
+    Random random = new Random(0);
+    int total = 0;
+    for (int i = 0; i < 10000; i++) {
+      random.nextBytes(bytes);
+      total += encodeDecode(new String(bytes, StandardCharsets.ISO_8859_1));
+    }
+    assertEquals(4190044,total);
```
srowen, line 324: Nit, here and elsewhere - space after comma

core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
```diff
@@ -175,4 +181,197 @@ public void testSampleWithNoDataNoMacro() throws FormatException {
...
+    encodeDecode("ABCD", 6);
+    encodeDecode("ABCDE", 4);
+    encodeDecode("ABCDEF", 4);
+    encodeDecode("ABCDEFG", 5);
+    encodeDecode("ABCDEFGH", 5);
+  }
+
+  @Test
+  public void testNumeric() throws WriterException, FormatException {
+    encodeDecode("1", 2);
```
srowen, line 199: I suppose you could write these type of tests with a loop, using substring of the entire string from 0 to N

PR #1506, opened 2022-03-04
Add support for multi-eci encoding for PDF417

AlexGeller1:
By default multi-ECI encoding is disabled so that the encoding behavior is identical with the previous version. 
The implementation shares the classes that convert character input to a byte/ECI stream with the data-matrix encoder.
This PR also fixes an issue that the PDF417 decoder incorrectly decoded input with a leading ECI and no explicit latch to TEXT encoding.

### DIFF
--- a/core/src/main/java/com/google/zxing/common/MinimalECIInput.java
+++ b/core/src/main/java/com/google/zxing/common/MinimalECIInput.java
@@ -1,0 +1,328 @@
+/*
+ * Copyright 2021 ZXing authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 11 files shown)


srowen:
The default behavior stays the same after this change?

srowen:
Have a look at the code alerts, probably really minor stuff

AlexGeller1:
Yes, thanks. Should be good now.

core/src/main/java/com/google/zxing/common/MinimalECIInput.java
```diff
@@ -0,0 +1,325 @@
...
+  *
+  * @param stringToEncode the character string to encode
+  * @param priorityCharset The preferred {@link Charset}. When the value of the argument is null, the algorithm
+  *   chooses charsets that leads to a minimal representation. Otherwise the algorithm will use the priority
+  *   charset to encode any character in the input that can be encoded by it if the charset is among the
+  *   supported charsets.
+  * @param fnc1 denotes the character in the input that represents the FNC1 character or -1 if this is not GS1
+  *   input.
+  */
+  public MinimalECIInput(String stringToEncode, Charset priorityCharset, int fnc1) {
```
srowen, line 47: Does the new ECI stuff need to be in common - is it generic to at least two barcode types?

AlexGeller1: Yes, it is common to data-matrix and PDF417.

AlexGeller1: It used to be an inner class of of datamatrix/encoder/MinimalEncoder.java

PR #1507, opened 2022-03-07
Add multi-eci decoding for PDF417

AlexGeller1:
Also fixed issue that some multi-eci encoded PDF417 codes were missing ECIs
The unit test now also decodes the multi-eci encoding tests and verifies that the result is identical with the input.

### DIFF
--- a/core/src/main/java/com/google/zxing/common/MinimalECIInput.java
+++ b/core/src/main/java/com/google/zxing/common/MinimalECIInput.java
@@ -208,6 +208,7 @@ public class MinimalECIInput implements ECIInput {
     return bytes[index] - 256;
   }
 
+  @Override
   public String toString() {
     StringBuilder result = new StringBuilder();
     for (int i = 0; i < length(); i++) {
--- a/core/src/main/java/com/google/zxing/common/MinimalECIInput.java
+++ b/core/src/main/java/com/google/zxing/common/MinimalECIInput.java
@@ -208,6 +208,26 @@ public class MinimalECIInput implements ECIInput {
     return bytes[index] - 256;
   }
 
+  public String toString() {
+    StringBuilder result = new StringBuilder();
+    for (int i = 0; i < length(); i++) {
+      if (i > 0) {
+        result.append(", ");
+      }
+      if (isECI(i)) {
+        result.append("ECI(");
+        result.append(getECIValue(i));
+        result.append(')');
+      } else if (charAt(i) < 128) {
+        result.append('\'');
+        result.append(charAt(i));
+        result.append('\'');
+      } else {
+        result.append((int) charAt(i));
+      }
+    }
+    return result.toString();
+  }
   static void addEdge(InputEdge[][] edges, int to, InputEdge edge) {
     if (edges[to][edge.encoderIndex] == null ||
         edges[to][edge.encoderIndex].cachedTotalSize > edge.cachedTotalSize) {
--- a/core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
+++ b/core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
@@ -22,10 +22,7 @@ import com.google.zxing.common.DecoderResult;
 import com.google.zxing.pdf417.PDF417ResultMetadata;
 
 import java.io.ByteArrayOutputStream;
-import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
-import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 
 /**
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/google/zxing/common/MinimalECIInput.java
```diff
@@ -208,6 +208,26 @@ public int getECIValue(int index) {
     return bytes[index] - 256;
   }
 
+  public String toString() {
```
srowen, line 211: Add `@Override` annotation

core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
```diff
@@ -749,4 +750,66 @@ private static String decodeBase900toBase10(int[] codewords, int count) throws F
     return resultString.substring(1);
   }
 
+  private static final class ECIOutput {
+    boolean needFlush = false;
```
srowen, line 754: These should be private?

core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
```diff
@@ -749,4 +750,66 @@ private static String decodeBase900toBase10(int[] codewords, int count) throws F
...
+      }
+      encodingName = charsetECI.name();
+    }
+
+    private void flush() {
+      if (needFlush) {
+        needFlush = false;
+        try {
+          result.append(bytes.toString(encodingName));
+        } catch (java.io.UnsupportedEncodingException uee) {
```
srowen, line 798: Nit: import class

core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
```diff
@@ -749,4 +750,66 @@ private static String decodeBase900toBase10(int[] codewords, int count) throws F
...
+          throw new IllegalStateException(uee);
+        }
+      }
+    }
+
+    private boolean isEmpty() {
+      return !needFlush && result.length() == 0;
+    }
+
+    public String toString() {
```
srowen, line 809: Likewise Override annotation

srowen (changes requested):
Minor comments.
Seems OK if this strictly increases the number of symbols that can be decoded

AlexGeller1:
>Seems OK if this strictly increases the number of symbols that can be decoded
Great. Yes, it does just that and it also fixes a bug in the encoding. Regarding the risks of this PR, I have the following thoughts:

- Regarding the encoding bug fix. The fix is in the function PDF417HighLevelEncoder.encodeMultiECIBinary(). This function is only called if the encoding hint  PDF417_AUTO_ECI is set so that the default behavior is not affected.
- DecodedBitStreamParser line 105 .. 113: The removed code was newly introduced in the previous PR (#1506) to fix the issue that the decoder was failing to decode a single ECI code with no leading latch. The fix is no longer needed since the method textCompaction now handles ECIs. In other words, the change reverts the code to the previous state.
- Changes to DecodedBitStreamParser.textCompaction(): The function now handles ECI_CHARSET in the switch case. There is no other behavior change. Before the change the ECI was ignored. As long as the new treatment doesn't crash, it may in the worse case replaces one buggy behavior with another.
- Changes to DecodedBitStreamParser.decodeTextCompaction(): There is no behavioral change. The method can now be instructed to start in a particular state. In the default case (no ECI) it starts with the original state Mode.ALPHA (see line 286). The method now returns the last state that it latched to.

AlexGeller1:
Please don't merge it yet. The spec says that ECIs may occur in the middle of 901 data (less than 6 bytes) and at the boundaries of 924 blocks (every 5 codewords) without requiring a new latch. The encoder never creates that but I am making sure that the decoder can handle it.

AlexGeller1:
I condensed the code of the function DecodedBitStreamParser.byteCompaction() so that it is about half as long as before.
I think that it is pretty straightforward now.

core/src/test/java/com/google/zxing/pdf417/decoder/PDF417DecoderTestCase.java
```diff
@@ -281,55 +281,68 @@ public void testBinaryData() throws WriterException, FormatException {
...
     //single ECI since the German Umlaut is in ISO-8859-1
-    performECITest(new char[] {'a', '1', '\u00c4', '\u042f'}, new float[] {20f, 1f, 1f, 10f}, 80529, 96007);
+    performECITest(new char[] {'a', '1', '\u00c4', '\u042f'}, new float[] {20f, 1f, 1f, 10f}, 80755, 96007);
   }
 
   @Test
   public void testECIEnglishCzechCyrillic1() throws Exception {
     //multi ECI between ISO-8859-2 and ISO-8859-5
-    performECITest(new char[] {'a', '1', '\u010c', '\u042f'}, new float[] {10f, 1f, 10f, 10f}, 91482, 124525);
+    performECITest(new char[] {'a', '1', '\u010c', '\u042f'}, new float[] {10f, 1f, 10f, 10f}, 102824, 124525);
```
srowen, line 320: Do you know the nature of the change in the expect result - probably fine? just looking for any points where this changes behavior. If it's correct and better behavior that's OK

AlexGeller1: Yes, this is the multi-ECI encoding test. This tests code that is only run if the encoding hint PDF417_AUTO_ECI is set.
The difference in size is due to the bug in multi-eci encoding that is also fixed in this PR as explained above (change in line 440 in PDF417HighLevelEncoder.encodeMultiECIBinary()).

srowen (approved):

Issue #1163, opened 2019-05-03
IndexOutOfBounds when encoding a PDF417


Doophie:
I'm trying to encode a PDF417 and whenever I set the `PDF417_COMPACTION` to `Compaction.TEXT` I get this crash: 

```
    java.lang.ArrayIndexOutOfBoundsException: length=128; index=8226
        at com.google.zxing.pdf417.encoder.PDF417HighLevelEncoder.isMixed(PDF417HighLevelEncoder.java:456)
        at com.google.zxing.pdf417.encoder.PDF417HighLevelEncoder.encodeText(PDF417HighLevelEncoder.java:298)
        at com.google.zxing.pdf417.encoder.PDF417HighLevelEncoder.encodeHighLevel(PDF417HighLevelEncoder.java:185)
        at com.google.zxing.pdf417.encoder.PDF417.generateBarcodeLogic(PDF417.java:649)
        at com.google.zxing.pdf417.PDF417Writer.bitMatrixFromEncoder(PDF417Writer.java:107)
        at com.google.zxing.pdf417.PDF417Writer.encode(PDF417Writer.java:87)
        at com.google.zxing.MultiFormatWriter.encode(MultiFormatWriter.java:102)
```

Here is the code for how I am trying to create the barcode: 

```
        var hints: MutableMap<EncodeHintType, Any>?
        
        hints = EnumMap<EncodeHintType, Any>(EncodeHintType::class.java)
        
        hints[EncodeHintType.CHARACTER_SET] = "UTF-8" 
        hints[EncodeHintType.ERROR_CORRECTION] = 6
        hints[EncodeHintType.PDF417_COMPACTION] = Compaction.TEXT
        
        val writer = MultiFormatWriter()
        val result: BitMatrix
        try {
            result = writer.encode(contents, format, img_width, img_height, hints)
        } catch (iae: WriterException) {
            // Unsupported format
            return null
        }
```

Is this a bug or am i doing something wrong?

srowen:
I suspect it's a bug, and I unfortunately have no idea what. I think there are a few latent bugs in the PDF417 parts. If you figure it out, feel free to open a PR. What language is this BTW?

Doophie:
will do, the language is kotlin

Doophie:
@srowen After looking into it a bit more, my data contained a • character and it was checking that character that it crashed and got the index out of bounds. If i have time later i'll see if i can fix the crash.

hidethatbug:
[...]

PR #1508, opened 2022-03-11
Share the ECI string builder of the data matrix decoder with the PDF417 decoder

AlexGeller1:
Besides the benefit of code reuse, the shared class is also more efficient in the no-ECI case so that the PDF417 decoder profits from that. With this PR we now have both a multi-eci encoder (..common.MinimalECIInput) and decoder (..common.ECIStringBuilder) that are shared between data-matrix and PDF417. I didn't look, but there may be other bar codes that don't yet support multi-eci, that could also make use of it.

### DIFF
--- a/core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
+++ b/core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2021 ZXing authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.zxing.common;
+
+import com.google.zxing.FormatException;
+
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Class that converts a sequence of ECIs and byte into a string to be used for decoding multi-eci content
+ *
+ * @author Alex Geller
+ */
+
+public final class ECIStringBuilder {
+  private StringBuilder currentBytes;
+  private StringBuilder result;
+  private Charset currentCharset = StandardCharsets.ISO_8859_1;
+
+  public ECIStringBuilder() {
+    currentBytes = new StringBuilder();
+  }
+  public ECIStringBuilder(int initialCapacity) {
+    currentBytes = new StringBuilder(initialCapacity);
+  }
+
+  public void append(char value) {
+    currentBytes.append((char) (value & 0xff));
+  }
+
+  public void append(byte value) {
+    currentBytes.append((char) (value & 0xff));
+  }
+
+  public void append(String value) {
+    currentBytes.append(value);
+  }
+
+  /* short for {@code append("" + value)}*/
+  public void append(int value) {
+    append("" + value);
+  }
+
+  public void appendECI(int value) throws FormatException {
+    encodeCurrentBytesIfAny();
+    CharacterSetECI characterSetECI = CharacterSetECI.getCharacterSetECIByValue(value);
+    if (characterSetECI == null) {
+      throw FormatException.getFormatInstance(new RuntimeException("Unsupported ECI value " + value));
+    }
+    currentCharset = characterSetECI.getCharset();
+  }
+
+  private void encodeCurrentBytesIfAny() {
+    if (currentCharset == StandardCharsets.ISO_8859_1) {
+      if (currentBytes.length() > 0) {
+        if (result == null) {
+          result = currentBytes;
+          currentBytes = new StringBuilder();
+        } else {
+          result.append(currentBytes);
+          currentBytes.setLength(0);
+        }
+      }
+    } else if (currentBytes.length() > 0) {
+      byte[] bytes = new byte[currentBytes.length()];
+      for (int i = 0; i < bytes.length; i++) {
+        bytes[i] = (byte) (currentBytes.charAt(i) & 0xff);
+      }
+      currentBytes.setLength(0);
+      if (result == null) {
+        result = new StringBuilder(new String(bytes, currentCharset));
+      } else {
+        result.append(new String(bytes, currentCharset));
+      }
+    }
+  }
+
+  /**
+   * appends the characters from sb (unlike all other append methods of this class who append bytes)
+   */
+  public void appendCharacters(StringBuilder value) {
+    encodeCurrentBytesIfAny();
+    result.append(value);
+  }
+
+  /**
+   * returns the length of toString()
+   */
+  public int length() {
+    return toString().length();
+  }
+
+  public boolean isEmpty() {
+    return currentBytes.length() == 0 && (result == null || result.length() == 0);
+  }
+
+  @Override
+  public String toString() {
+    encodeCurrentBytesIfAny();
+    return result == null ? "" : result.toString();
+  }
+}
--- a/core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
+++ b/core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
@@ -39,28 +39,28 @@ public final class ECIStringBuilder {
   }
 
   /**
-   * appends {@code value} as a byte value
+   * Appends {@code value} as a byte value
    */
   public void append(char value) {
     currentBytes.append((char) (value & 0xff));
   }
 
   /**
-   * appends {@code value} as a byte value (not it's string representation)
+   * Appends {@code value} as a byte value (not its string representation)
    */
   public void append(byte value) {
     currentBytes.append((char) (value & 0xff));
   }
 
   /**
-   * appends the characters in {@code value} as bytes values
+   * Appends the characters in {@code value} as bytes values
    */
   public void append(String value) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Class that converts a sequence of ECIs and bytes into a string
+ *
+ * @author Alex Geller
+ */
+
```
srowen, line 29: Nit: remove blank line to connect the javadoc

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+   */
+  public void append(String value) {
+    currentBytes.append(value);
+  }
+
+  /**
+   * append the string repesentation of {@code value} (short for {@code append("" + value)})
+   */
+  public void append(int value) {
+    append("" + value);
```
srowen, line 67: String.valueOf(value)

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  public void appendECI(int value) throws FormatException {
+    encodeCurrentBytesIfAny();
+    CharacterSetECI characterSetECI = CharacterSetECI.getCharacterSetECIByValue(value);
+    if (characterSetECI == null) {
+      throw FormatException.getFormatInstance(new RuntimeException("Unsupported ECI value " + value));
```
srowen, line 74: Pass the message directly, no need for RuntimeException

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+    currentCharset = characterSetECI.getCharset();
+  }
+
+  private void encodeCurrentBytesIfAny() {
+    if (currentCharset == StandardCharsets.ISO_8859_1) {
```
srowen, line 80: .equals() not ==

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+          currentBytes = new StringBuilder();
+        } else {
+          result.append(currentBytes);
+          currentBytes.setLength(0);
+        }
+      }
+    } else if (currentBytes.length() > 0) {
+      byte[] bytes = new byte[currentBytes.length()];
+      for (int i = 0; i < bytes.length; i++) {
+        bytes[i] = (byte) (currentBytes.charAt(i) & 0xff);
```
srowen, line 93: This assumes chars are single-byte, is that true? turning the string into bytes according to some encoding with getBytes() sounds better (according to ISO-8859-1)

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+      }
+      currentBytes.setLength(0);
```
srowen, line 95: Reset to a new StringBuilder

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+      if (result == null) {
+        result = new StringBuilder(new String(bytes, currentCharset));
+      } else {
+        result.append(new String(bytes, currentCharset));
+      }
+    }
+  }
+
+  /**
+   * appends the characters from sb (unlike all other append methods of this class who append bytes)
```
srowen, line 105: Nit: capitalize 'appends' and does a caller know what sb is?

srowen (changes requested):

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
AlexGeller1: Are you sure? I see only these two public methods:
public static FormatException getFormatInstance()
public static FormatException getFormatInstance(Throwable cause)

srowen: Ah right, just leave out the msg (not super important)

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
AlexGeller1: Yes, very good.

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * appends {@code value} as a byte value
+   */
+  public void append(char value) {
+    currentBytes.append((char) (value & 0xff));
+  }
+
+  /**
+   * appends {@code value} as a byte value (not it's string representation)
```
srowen, line 49: Nit: it's -> its
Go ahead and capitalize append here and elsewhere

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+   */
+  public void append(byte value) {
+    currentBytes.append((char) (value & 0xff));
+  }
+
+  /**
+   * appends the characters in {@code value} as bytes values
+   */
+  public void append(String value) {
```
srowen, line 58: Internally, this isn't tracking this as bytes, but as a String. That's OK if that's the correct semantics but it seems at odds with the docs. At the least isn't it true that it will ultimately use the ISO-8859-1 representation of the strings? you could build up a byte array directly instead (ByteArrayOutputStream)

AlexGeller1: The outside semantics of the append methods is clearly bytes. But this class uses a StringBuilder as internal representation (and not, for example a ByteArray) because is used from the data matrix and the PDF417 decoders who both originally used StringBuilder (before the multi ECI support was added)  to decode the bytes. If we now use ByteArray here then we will degrade the performance because we will need a conversion from ByteArray to String at some point. The way that this class is written now, it has the same performance as the original StringBuilder in the case that there is no ECI (which is the most common case). 
The code that I have looked into in ZXing (code-128, qr-code, data-matrix and pdf417) uses Strings everywhere but in a large part of the code it is encoding and decoding bytes because these strings are assumed to not contain values greater 255. The differentiation may not have been important but this class now bridges between strings that only contain ISO-8859-1 characters (I call them bytes) and Strings that contain Unicode.  
By the way, I haven't looked at the details of issue #1163 but I suspect a missing check that would reject characters or trim Unicode characters to ISO-8859-1. Without this check or trim step, the PDF417 encoder assumes that the input contains only is ISO-8859-1 characters and it can for example produce an ArrayIndexOutOfBoundsException as described in the bug when a larger value is encountered.

srowen: I think I'm generally fine staying in Strings in this process, but to your comment, this class already round-trips from String -> byte[] -> String, so seems like fewer hops, not more. I am worried about conflating strings and byte[]s back and forth - recipe for character-encoding-related problems. More direct seems better to reason about. If it's for convenience -- caller already makes the same encoding assumption -- I could see it. Just don't want to use string (assumed to be ISO-8859-1) a replacement for byte[] as it's so easy to work in byte[]s where they're bytes.

AlexGeller1: There are fewer hops if we leave the code as is. The situation is as follows: The original code of both the data-matrix decoder and the PDF417 decoder declared "result" as StringBuilder() and appended ISO-8859-1 characters using the various StringBuilder append() methods. 
I then replaced "StringBuilder result=new StringBuilder()" with "ECIStringBuilder result=new ECIStringBuilder()" and the code remained largely the same since ECIStringBuilder provides the same append() methods. If the method ECIStringBuilder.appendECI() is not called during decoding then exactly on StringBuilder will have been constructed during the process and the final string will be obtained from it via result.toString() as in the original code before ECIs were introduced. There is never a conversion to bytes involved anywhere. Now, when result.appendECI() is called, then the ISO-8859-1 "characters" following the ECI are bytes of  the ECI that precedes them and will be converted to characters either on the next call to appendECI() or at latest when toString() is called. Here we have a character-byte-character conversion that could perhaps be changed but doing that without slowing the general case with no ECI will not be straightforward. The current code also avoids the byte-character conversion when appendECI() is called with a ISO-8859-1 ECI after if has previously been set to another encoding. This would not be possible if bytes were stored.

core/src/main/java/com/google/zxing/common/ECIStringBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+        if (result == null) {
+          result = currentBytes;
+          currentBytes = new StringBuilder();
+        } else {
+          result.append(currentBytes);
+          currentBytes = new StringBuilder();
+        }
+      }
+    } else if (currentBytes.length() > 0) {
+      byte[] bytes = currentBytes.toString().getBytes(StandardCharsets.ISO_8859_1);
```
srowen, line 90: Hm but there is a conversion back and forth here and line 93. Why can this not stay in strings?  I imagine there is indeed a good reason in the end, just checking my basic understanding

AlexGeller1: Yes, we get to line 93 if the test in line 79 is negative (testing and finding that the current encoding isn't ISO-8859-1). In this case we have to do a character-byte-character conversion. This happens when appendECI() is called with a non ISO-8859-1 value. In the general case this is not called and also in the case that there is a switchback to ISO-8859-1, we use the efficient path again.

srowen (approved):

PR #1514, opened 2022-04-19
Replacing non ISO-8859-1 characters in PDF417 input with '?'

AlexGeller1:
… when neither a character-set nor multi-eci-encoding is selected.
This fixes issue #1163.

### DIFF
--- a/core/src/main/java/com/google/zxing/pdf417/encoder/PDF417HighLevelEncoder.java
+++ b/core/src/main/java/com/google/zxing/pdf417/encoder/PDF417HighLevelEncoder.java
@@ -174,9 +174,9 @@ final class PDF417HighLevelEncoder {
     if (encoding == null && !autoECI) {
       for (int i = 0; i < msg.length(); i++) {
         if (msg.charAt(i) > 255) {
-          //will produce '?' for characters with a value greater than 255
-          msg = new String(msg.getBytes(StandardCharsets.ISO_8859_1), StandardCharsets.ISO_8859_1);
-          break;
+          throw new WriterException("Non-encodable character detected: " + msg.charAt(i) + " (Unicode: " +
+              (int) msg.charAt(i) +
+              "). Consider specifying EncodeHintType.PDF417_AUTO_ECI and/or EncodeTypeHint.CHARACTER_SET.");
         }
       }
     }
--- a/core/src/main/java/com/google/zxing/pdf417/encoder/PDF417HighLevelEncoder.java
+++ b/core/src/main/java/com/google/zxing/pdf417/encoder/PDF417HighLevelEncoder.java
@@ -171,6 +171,15 @@ final class PDF417HighLevelEncoder {
   static String encodeHighLevel(String msg, Compaction compaction, Charset encoding, boolean autoECI) 
       throws WriterException {
 
+    if (encoding == null && !autoECI) {
+      for (int i = 0; i < msg.length(); i++) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/google/zxing/pdf417/encoder/PDF417HighLevelEncoder.java
```diff
@@ -171,6 +171,15 @@ private PDF417HighLevelEncoder() {
   static String encodeHighLevel(String msg, Compaction compaction, Charset encoding, boolean autoECI) 
       throws WriterException {
 
+    if (encoding == null && !autoECI) {
+      for (int i = 0; i < msg.length(); i++) {
+        if (msg.charAt(i) > 255) {
+          //will produce '?' for characters with a value greater than 255
+          msg = new String(msg.getBytes(StandardCharsets.ISO_8859_1), StandardCharsets.ISO_8859_1);
```
srowen, line 178: I don't think this replaces with "?" - I tried a simple example. The results will be kinda undefined; you may get a "?" for unprintable chars in the console.
https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetDecoder.html will do this more correctly, I think, as it lets you present a replacement char.

Should we just reject this input? encoding some other string doesn't sound right.
These chars just are not representable at all in PDF417?

AlexGeller1: >I don't think this replaces with "?" - I tried a simple example. The results will be kinda undefined; you may get a "?" for unprintable chars in the console.

On my machine I just tried this:
```java
for(int i=256;i<0xffff;i++)
{
    System.err.println("DEBUG i="+i+", val="+(""+(char)i).getBytes(java.nio.charset.StandardCharsets.ISO_8859_1)[0]);
}
```
which yieds:
```code
DEBUG i=256, val=63
DEBUG i=257, val=63
...
DEBUG i=65533, val=63
DEBUG i=65534, val=63
```
But never mind, as you point out, the question is how it should behave.
We could take the point of view that the input, which should be by definition ISO-8859-1, is decoded/encoded using the standard Java ISO-8859-1 encoder/decoder.
>https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetDecoder.html will do this more correctly, I think, as it lets you present a replacement char.

Personally I would prefer to do whatever the standard decoder does but it you like, yes I can use that instead. 
>Should we just reject this input? encoding some other string doesn't sound right.

Absolutely, I thought about that but it may not be backward compatible. I am not sure that you get issue #1163 on any input > 255 since the array access is only done in a specific encoding mode. If by chance another encoding mode is chosen then it may be that there will be no runtime exception. If we now raise an error then this would be a behavior change.
>These chars just are not representable at all in PDF417?

Exactly, they are not representable without an ECI.

srowen: If it's supposed to be ISO-8859-1, I would throw a WriterException in this case, fail more explicitly.

AlexGeller1:
Ping in case that you didn't see commit 4d2367e6f37a7c8d6f5977b3fded61d49521106b (I marked the conversation as resolved before the commit so that I am thinking that perhaps you weren't notified. I you were, then no worries. I am not pushing).

srowen (approved):

PR #1516, opened 2022-04-28
Specify `android:exported` to the activities having `<intent-filter>`

PatilShreyas:
While targetting to Android 12 SDK (API 31), there's a restriction to providing the flag `android:exported` to all the components (activity, service, receivers) having `<intent-filter>`. This PR sets this flag to all activities having intent filters which can be helpful for the client applications which are targeting to Android 12.

Closes: #1517

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -48,7 +48,8 @@
               android:clearTaskOnLaunch="true"
               android:stateNotNeeded="true"
               android:theme="@style/CaptureTheme"
-              android:windowSoftInputMode="stateAlwaysHidden">
+              android:windowSoftInputMode="stateAlwaysHidden"
+              android:exported="false">
       <intent-filter>
         <action android:name="android.intent.action.MAIN"/>
         <category android:name="android.intent.category.LAUNCHER"/>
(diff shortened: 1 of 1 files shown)


srowen:
I doubt this is all the change that is needed? IIRC the permissions model has to change too. Does it work?

PatilShreyas:
Yeah this change is needed. Otherwise consumer apps will manually need to override entries in the manifest by exporting it as false and merging nodes.

srowen:
Is it sufficient though to make the app work? I think there are other changes needed. The change itself seems fine, just not exporting the activities

PatilShreyas:
As far as I saw, this change seems fine. What do you think what kind of other changes are needed which are mentioned [here](https://developer.android.com/about/versions/12/behavior-changes-12)?

PatilShreyas:
But as you can see, I haven't updated SDK target or compile version. I've just specified `exported` flags. I don't thing it would cause any issues.

srowen:
Right. I am saying that this is likely insufficient for Android 12 compatibility. Change the title?

PatilShreyas:
So the thing is app is unable to compile when this library is included in app which is targetting android 12. Manifest issues are occurring because this library haven't specified this flag (_exported_). So this change will make sure that no compile time issues will occur whenever any app is using it. Otherwise, every app will need to override Manifest entries at their end manually.

PatilShreyas:
Cool, I've updated PR title

PR #1522, opened 2022-05-04
Adds orientation of pdf417 barcode to Result metadata

wildjordanpaul:
Pull Request https://github.com/zxing/zxing/pull/1333 added support for scanning PDF417 barcodes at different orientations. We have a use case where we are using the orientation of the barcode to know the correct orientation of the page. With previous versions of zxing, we attempted to read barcodes from images with each orientation. With 3.5.0, the different orientations are now scanned in the Detector, with no way of the client to know which orientation the barcode was found in. This change stores the orientation in the result's metadata so that the orientation data can be used by clients of this library.

I also modified the rotation logic such that we don't clone the matrix when decoding with a rotation of 0.

### DIFF
--- a/core/src/main/java/com/google/zxing/pdf417/detector/Detector.java
+++ b/core/src/main/java/com/google/zxing/pdf417/detector/Detector.java
@@ -79,7 +79,7 @@ public final class Detector {
     //boolean tryHarder = hints != null && hints.containsKey(DecodeHintType.TRY_HARDER);
 
     BitMatrix originalMatrix = image.getBlackMatrix();
-    for(int rotation : ROTATIONS) {
+    for (int rotation : ROTATIONS) {
       BitMatrix bitMatrix = applyRotation(originalMatrix, rotation);
       List<ResultPoint[]> barcodeCoordinates = detect(multiple, bitMatrix);
       if (!barcodeCoordinates.isEmpty()) {
@@ -98,11 +98,11 @@ public final class Detector {
   private static BitMatrix applyRotation(BitMatrix matrix, int rotation) {
     if (rotation % 360 == 0) {
       return matrix;
-    } else {
-      BitMatrix newMatrix = matrix.clone();
-      newMatrix.rotate(rotation);
-      return newMatrix;
     }
+
+    BitMatrix newMatrix = matrix.clone();
+    newMatrix.rotate(rotation);
+    return newMatrix;
   }
 
   /**
--- a/core/src/main/java/com/google/zxing/pdf417/detector/Detector.java
+++ b/core/src/main/java/com/google/zxing/pdf417/detector/Detector.java
@@ -57,6 +57,7 @@ public final class Detector {
   // ensure we don't miss it.
   private static final int ROW_STEP = 5;
   private static final int BARCODE_MIN_HEIGHT = 10;
+  private static final int[] ROTATIONS = {0, 180, 270, 90};
 
   private Detector() {
   }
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/zxing/pdf417/detector/Detector.java
```diff
@@ -77,20 +78,31 @@ public static PDF417DetectorResult detect(BinaryBitmap image, Map<DecodeHintType
...
-    List<ResultPoint[]> barcodeCoordinates = detect(multiple, bitMatrix);
-    // Try 180, 270, 90 degree rotations, in that order
-    for (int rotate = 0; barcodeCoordinates.isEmpty() && rotate < 3; rotate++) {
-      bitMatrix = bitMatrix.clone();
-      if (rotate != 1) {
-        bitMatrix.rotate180();
-      } else {
-        bitMatrix.rotate90();
+    BitMatrix originalMatrix = image.getBlackMatrix();
+    for(int rotation : ROTATIONS) {
```
srowen, line 82: Looks good. The style checker is complaining about the lack of space after `for`: `src/main/java/com/google/zxing/pdf417/detector/Detector.java:[82,5] (whitespace) WhitespaceAround: 'for' is not followed by whitespace.`

core/src/main/java/com/google/zxing/pdf417/detector/Detector.java
```diff
@@ -77,20 +78,31 @@ public static PDF417DetectorResult detect(BinaryBitmap image, Map<DecodeHintType
...
+  /**
+   * Applies a rotation to the supplied BitMatrix.
+   * @param matrix bit matrix to apply rotation to
+   * @param rotation the degrees of rotation to apply
+   * @return BitMatrix with applied rotation
+   */
+  private static BitMatrix applyRotation(BitMatrix matrix, int rotation) {
+    if (rotation % 360 == 0) {
+      return matrix;
+    } else {
```
srowen, line 101: Total nit, just omit the else { } statement and leave the block as the remainder of the method

srowen (changes requested):

core/src/main/java/com/google/zxing/pdf417/detector/Detector.java
wildjordanpaul: Done, thanks!

core/src/main/java/com/google/zxing/pdf417/detector/Detector.java
wildjordanpaul: Not a problem, changed 👍

core/src/main/java/com/google/zxing/pdf417/detector/PDF417DetectorResult.java
```diff
@@ -28,10 +28,12 @@ public final class PDF417DetectorResult {
 
   private final BitMatrix bits;
   private final List<ResultPoint[]> points;
+  private final int rotation;
 
-  public PDF417DetectorResult(BitMatrix bits, List<ResultPoint[]> points) {
+  public PDF417DetectorResult(BitMatrix bits, List<ResultPoint[]> points, int rotation) {
```
srowen, line 33: `com.google.zxing.pdf417.detector.PDF417DetectorResult: In method 'public PDF417DetectorResult(com.google.zxing.common.BitMatrix, java.util.List)' the number of arguments has changed` Ah yeah it has a point. Let's keep the original constructor as well and have it default to rotation=0

wildjordanpaul: Oh yeah, great point!

srowen (approved):

jackromo888 (commented):
Thannks

Issue #1427, opened 2021-09-01
Java Exception: java.lang.IllegalStateException: Failed to recover - encoded with PDF_417 in 110x166


Microsvuln:
Hi .

Using jazzer, the following exception has been occurred :

`Java Exception: java.lang.IllegalStateException: Failed to recover` .

The produced jazzer file is as :
```

import java.lang.reflect.InvocationTargetException;
import java.lang.reflect.Method;
public class Crash_554492f3abaea0e383717bff996d44d0cdb12f5f {
    static final String base64Bytes = "rO0ABXNyABNqYXZhLnV0aWwuQXJyYXlMaXN0eIHSHZnHYZ0DAAFJAARzaXpleHAAAAADdwQAAAADc3IAEWphdmEubGFuZy5JbnRlZ2VyEuKgpPeBhzgCAAFJAAV2YWx1ZXhyABBqYXZhLmxhbmcuTnVtYmVyhqyVHQuU4IsCAAB4cAAAAG5zcQB+AAIAAACAdAAAeA==";

    public static void main(String[] args) {
        ClassLoader.getSystemClassLoader().setDefaultAssertionStatus(true);
        try {
            Method fuzzerInitialize = MultiFormatEncodeFuzzer.class.getMethod("fuzzerInitialize");
            fuzzerInitialize.invoke(null);
        } catch (NoSuchMethodException ignored) {
            try {
                Method fuzzerInitialize = MultiFormatEncodeFuzzer.class.getMethod("fuzzerInitialize", String[].class);
                fuzzerInitialize.invoke(null, (Object) args);
            } catch (NoSuchMethodException ignored) {
            } catch (IllegalAccessException | InvocationTargetException e) {
                e.printStackTrace();
                System.exit(1);
            }
        } catch (IllegalAccessException | InvocationTargetException e) {
            e.printStackTrace();
            System.exit(1);
        }
        com.code_intelligence.jazzer.api.CannedFuzzedDataProvider input = new com.code_intelligence.jazzer.api.CannedFuzzedDataProvider(base64Bytes);
        MultiFormatEncodeFuzzer.fuzzerTestOneInput(input);
    }
}


```
And the corpus required to reproduce (Please first unzip it) :

[crash-corpus.zip](https://github.com/zxing/zxing/files/7092192/crash-corpus.zip)

To reproduce :

`./MultiFormatEncodeFuzzer  [crashing-corpus]
`
Thanks!

srowen:
[...]

PR #1523, opened 2022-05-05
Prevent encoding of the empty string for PDF417

AlexGeller1:
This fixes issue #1427

### DIFF
--- a/core/src/test/java/com/google/zxing/pdf417/encoder/PDF417EncoderTestCase.java
+++ b/core/src/test/java/com/google/zxing/pdf417/encoder/PDF417EncoderTestCase.java
@@ -16,6 +16,8 @@
 
 package com.google.zxing.pdf417.encoder;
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/zxing/pdf417/encoder/PDF417EncoderTestCase.java
```diff
@@ -67,4 +69,12 @@ public void testEncodeByte() throws Exception {
     assertEquals("\u039f\u001A\u0385abcd", encoded);
   }
 
+  @Test
```
srowen, line 72: Make this `@Test(expected=WriterException.class)` and you can remove the try-catch

AlexGeller1: Nice

srowen (approved):

PR #1570, opened 2022-11-04
Updated CaptureActivity

vtangade:
To set Set Zoom on Camera


srowen:
No, don't think this should be exposed

vtangade:
Okay, Can we add something to set Zoom on Camera please ?

Thank you

PR #1595, opened 2023-02-10
fix: 1358 - rolled back to old-fashioned Collections.sort for android 23

monsieurtanuki:
cf. https://developer.android.com/reference/java/util/List#sort(java.util.Comparator%3C?%20super%20E%3E), "added in API level 24"

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
+++ b/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
@@ -25,6 +25,7 @@ import com.google.zxing.common.BitMatrix;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
@@ -628,7 +629,10 @@ public class FinderPatternFinder {
       }
     }
 
-    possibleCenters.sort(moduleComparator);
+    // A more up-to-date version would be "possibleCenters.sort(moduleComparator);"
+    // But we need this old syntax for android API 23 (Marshmallow) and below
+    // cf. https://github.com/zxing/zxing/issues/1358
+    Collections.sort(possibleCenters, moduleComparator);
 
     double distortion = Double.MAX_VALUE;
     FinderPattern[] bestPatterns = new FinderPattern[3];


srowen:
No, not going to do this. Java 7 is just too old to contemplate supporting and I don't think this change alone makes it work on older versions. You can just use previous versions of the lib if you must

monsieurtanuki:
> No, not going to do this. Java 7 is just too old to contemplate supporting

I'm not asking you to support Java 7, it's about using an old but valid syntax. Actually there are other places in your code that also use this old syntax.

> I don't think this change alone makes it work on older versions.

I think your assumptions are wrong, but I'm not going to delve into how android works with java libraries.

> You can just use previous versions of the lib if you must

I can't, because I'm using a lib that uses zxing. That would mean I would have to copy the whole lib (and downgrade it to use an old zxing version). So, indeed technically I can, but that would be a bit in contradiction with the concept of libraries, and I wouldn't have the latest zxing bug fixes.

monsieurtanuki:
@srowen Feel free to close this PR if relevant.

monsieurtanuki:
@srowen Final answer?

srowen:
It is not the only thing that's needed - see like N other discussions. Just not going to support Java 6/7 anymore; people that want to support really old Android can use a prior version or do what they want in a fork

monsieurtanuki:
> Hm, I don't think old Android versions support Java 8 bytecode, as I recall.

https://developer.android.com/studio/write/java8-support?hl=en#library-desugaring
...If you're building your app using the Android Gradle plugin 4.0.0 or higher, the plugin extends support for using a number of Java 8 language APIs without requiring a minimum API level for your app.
This additional support for older platform versions is possible because plugin 4.0.0 and higher extends the desugaring engine to also desugar Java language APIs. You can include standard language APIs that were available only in recent Android releases (such as java.util.streams) in apps that support older versions of Android...

> What is y'all's use case for targeting something that doesn't support 8? apps that old won't even install anymore on Play

Oh yes they do install, on my own smartphone.

srowen:
OK, so you're compiling this code in your build. I'm more worried about the published library and adding that as a dependency, I thought that didn't work if you pulled in a Java 8 dependency on old android. Does dex handle all the desugaring? 

Yeah I know you can side-load it, I know that Play doesn't even let you publish such an app targeting old Android anyway and won't even let users download it recently, so just wonder how much this is actually a use case. Maybe this is for private apps?

OK if it's just this one change I suppose I am not super against it, if people really find it useful

monsieurtanuki:
> OK, so you're compiling this code in your build. I'm more worried about the published library and adding that as a dependency, I thought that didn't work if you pulled in a Java 8 dependency on old android. Does dex handle all the desugaring?

The beauty of it is that it is transparent for both of us - your java library is still in Java 8, and the desugaring takes place in android through minor instructions on the `build.gradle` compile file.
Actually I already use your Java 8 library on my android 21 Lollipop.
The only problem being with that particular syntax in one line, about QR codes, that the current PR aims to "explicitly desugar".

> Yeah I know you can side-load it, I know that Play doesn't even let you publish such an app targeting old Android anyway and won't even let users download it recently, so just wonder how much this is actually a use case. Maybe this is for private apps?

I have a published app that supports android 16+ (Jelly Bean) and I uploaded a new app version a couple of months ago to Google Play.

There may be confusion because of messages like that one: "New apps and app updates must target API level 33".
There are `compileSdkVersion`, `minSdkVersion` and `targetSdkVersion`:
* typically, `compileSdkVersion` and `targetSdkVersion` are identical, and Google Play forces you to upgrade every year to almost the latest version (in my published app it's still `31` and at the end of the year I'll have to update at least to `32`) (current max value is `33` I think)
* `minSdkVersion` (in my published app it's `16`)

> OK if it's just this one change I suppose I am not super against it, if people really find it useful

Thank you so much!

PR #1657, opened 2023-06-20
Add ResultMetadataType.ERRORS_CORRECTED and ERASURES_CORRECTED

gredler:
This change set adds metadata for `ERRORS_CORRECTED` and `ERASURES_CORRECTED` to `ResultMetadataType`.

`ERRORS_CORRECTED` is available for QR Code, PDF417, Maxicode, Data Matrix, and Aztec Code.

`ERASURES_CORRECTED` is available for PDF417.

Tests are included for all affected symbologies.

Error correction information can be useful to gauge the quality of the input image, or to verify that the encoding system was error-free.

For example, in my specific case I'd like to make sure that any barcodes generated by [Okapi](https://github.com/woo-j/OkapiBarcode/) can be decoded without requiring any error correction, to make sure there are no hidden encoding bugs.


core/src/main/java/com/google/zxing/aztec/AztecDetectorResult.java
```diff
@@ -31,16 +31,19 @@ public final class AztecDetectorResult extends DetectorResult {
...
   private final int nbDatablocks;
   private final int nbLayers;
+  private final int errorsCorrected;
 
   public AztecDetectorResult(BitMatrix bits,
                              ResultPoint[] points,
                              boolean compact,
                              int nbDatablocks,
-                             int nbLayers) {
+                             int nbLayers,
```
srowen, line 40: Retain the existing constructor - easy, and just ensures there is no binary compatibility issue.

srowen (changes requested):

core/src/main/java/com/google/zxing/aztec/AztecDetectorResult.java
gredler: Done. I also changed one of the test classes back to using the new (old) constructor, so that both constructors are exercised during the tests.

srowen:
OK. We just have ` 7006: com.google.zxing.common.reedsolomon.ReedSolomonDecoder: Return type of method 'public void decode(int[], int)' has been changed to int` now. Just because of how looong all these methods have been public, I'd say add a new differently-named method that returns an int, and have the old method call it and ignore the return value.

gredler:
Barely 16 years :-) Do you have any suggestions for the method name? I'd be tempted to use `decode2`, but the existing codebase skews towards more descriptive names, generally. Some other options I can think of: `decodeAndGetErrorsCorrected`, `decodeToInt`, `decodeWithCount`, `decodeAndCount`.

srowen:
deocdeWithECCount or something seems not terrible. Doesn't really matter.

PR #1658, opened 2023-06-23
Allow space in Code128 CodeSet B

primo-ppcg:
Space is a valid character in CodeSet B.


srowen (approved):
Hm! yes I think that's right.

Issue #1672, opened 2023-08-03
Code128 Barcode - Maximum Limit of content is 80 chars


himanshushukla254:
Hi Team, I was going through the code128 writer present in `com.google.zxing.oned` package and found that the max length of allowed content is set to 80.

Any reasons for setting this limit? I see code128 algo does not have any limits implicitly and other online platforms are also generating Code128 barcode with more than 80 chars.

Attached Screenshot from Code128Writer:

<img width="632" alt="image" src="https://github.com/zxing/zxing/assets/15158684/daeadfb5-0ffd-4394-9d94-dfc61f582605">

srowen:
I'm not sure. Do you know if there is a max in any spec? If there's no evidence there is really a limit we could remove this, but there are practical upper limits for sure.

himanshushukla254:
Not sure, There are different statements in many articles regarding the max-content length.
https://www.softmatic.com/barcode-code-128.html#128-max
https://stackoverflow.com/questions/24263237/what-is-the-maximum-length-of-a-code-128-barcode

Even this platform also supporting >80 chars
https://barcodeapi.org/index.html#128

gredler:
For what it's worth, I had a look at my copy of the Code 128 spec (ISO/IEC 15417:2007) and don't see any formal maximum. I suspect this is more in the way of a sanity check (as far as my own experience goes, I don't think I've ever seen a Code 128 barcode longer than 50 characters or so in the wild).

srowen:
I think it's fine to take out this limit, but, in practice you won't really be able to use longer barcodes well anyway

himanshushukla254:
Yes, we can remove this forced check. This looks like a limitation in zxing as compared to other platforms.

himanshushukla254:
The changes are already done for this. This will be available in the 3.5.3 release as per [#1675 ]. Hence, Closing this issue.

PR #1675, opened 2023-08-16
Removing the forced content limit check for Code128Writer.java

himanshushukla254:
Removing the forced content limit check as per https://github.com/zxing/zxing/issues/1672


core/src/main/java/com/google/zxing/oned/Code128Writer.java
```diff
@@ -82,13 +82,6 @@ public boolean[] encode(String contents, Map<EncodeHintType,?> hints) {
   }
 
   private static int check(String contents, Map<EncodeHintType,?> hints) {
-    int length = contents.length();
```
srowen, line 85: Don't remove this, it's used later. (You can move it down if you like)

srowen (approved):

Issue #1682, opened 2023-09-07
CommandLineRunner --raw option is broken for PDF_417 barcodes


dlenski:
PDF_417 barcodes (and maybe other types as well) will break the `--raw` option, apparently because their `com.google.zxing.Result` method `getRawBytes()` returns `null`.

---

Using v3.5.1 and the [test/barcodes/PDF_417-easy.bmp](https://raw.githubusercontent.com/dlenski/python-zxing/master/test/barcodes/PDF_417-easy.bmp) from https://github.com/dlenski/python-zxing as an example:

```bash
$ java -cp 'core.jar:javase.jar:jcommander.jar' com.google.zxing.client.j2se.CommandLineRunner PDF_417-easy.bmp --raw
Exception in thread "main" java.lang.NullPointerException
	at com.google.zxing.client.j2se.DecodeWorker.decode(DecodeWorker.java:172)
	at com.google.zxing.client.j2se.DecodeWorker.call(DecodeWorker.java:76)
	at com.google.zxing.client.j2se.CommandLineRunner.main(CommandLineRunner.java:99)
```

The error is coming from here:
https://github.com/zxing/zxing/blob/557b5fcf0e22a6cf2541a54bd830116712fec51d/javase/src/main/java/com/google/zxing/client/j2se/DecodeWorker.java#L172-L175

The `--raw` option was added in https://github.com/zxing/zxing/commit/83650ce341732b5835c08411c8629a94a0fab082

PR #1683, opened 2023-09-07
Prevent --raw CLR option from failing if result.getRawBytes() is null

dlenski:
Fixes https://github.com/zxing/zxing/issues/1682


srowen (approved):

srowen:
Looks good, thanks!

PR #1693, opened 2023-09-15
Avoid decoding stacked row 1 of N independently

gredler:
When processing RSS Expanded stacked symbols, in some rare cases the first row processed looks like an independent single-row barcode based on checksum calculations. However, this is obviously not the case when the allowed finder pattern sequences are taken into account.

This PR adds a check to the finder pattern sequence as a last step before deciding to return the current (first processed) row as a fully, independently decoded barcode. I believe this also addresses a long-standing TODO in the code, now removed as part of this PR.

This PR includes a test which fails without the fix.


srowen (approved):

core/src/main/java/com/google/zxing/oned/rss/expanded/RSSExpandedReader.java
```diff
@@ -249,9 +248,10 @@ private List<ExpandedPair> checkRows(List<ExpandedRow> collectedRows, int curren
   }
 
   // Whether the pairs form a valid finder pattern sequence, either complete or a prefix
-  private static boolean isValidSequence(List<ExpandedPair> pairs) {
+  private static boolean isValidSequence(List<ExpandedPair> pairs, boolean complete) {
 
     for (int[] sequence : FINDER_PATTERN_SEQUENCES) {
+      boolean sizeOk = (complete ? pairs.size() == sequence.length : pairs.size() <= sequence.length);
```
srowen, line 254: This var isn't used, that's why codacy is complaining

gredler: Oops, missed a line. Fix incoming!

gredler:
@srowen I think this one's good now, let me know if you need anything else. Thanks!

PR #1726, opened 2023-11-27
Refactored some Java Files in this project as part of my University Assignment

jahnavi-prasad:
File Name: ResultParser.java
Refactoring Techniques used:
1. Rename method/variable
2. Decompose conditional
3. Introduce explaining variable
4. Move method/field

File Name: GUIRunner.java
Refactoring Techniques used: Extract class

File Name: VINResultParser.java
Refactoring Techniques used: Replace Conditional With Polymorphism

### DIFF
--- a/core/src/main/java/com/google/zxing/client/result/ResultParser.java
+++ b/core/src/main/java/com/google/zxing/client/result/ResultParser.java
@@ -29,7 +29,7 @@ import java.util.regex.Pattern;
 /**
  * <p>Abstract class representing the result of decoding a barcode, as more than
  * a String -- as some type of structured data. This might be a subclass which represents
- * a URL, or an e-mail address. {@link #parseResult(Result)} will turn a raw
+ * a URL, or an e-mail address. {@link #parseBarcodeResult(Result)} will turn a raw
  * decoded string into the most appropriate type of structured representation.</p>
  *
  * <p>Thanks to Jeff Griffin for proposing rewrite of these classes that relies less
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/zxing/client/result/ResultParser.java
```diff
@@ -74,27 +47,50 @@ public abstract class ResultParser {
...
       text = text.substring(1);
     }
     return text;
   }
 
-  public static ParsedResult parseResult(Result theResult) {
-    for (ResultParser parser : PARSERS) {
-      ParsedResult result = parser.parse(theResult);
+  public static ParsedResult parseBarcodeResult(Result barcodeResult) {
+    ResultParser[] parsers = {
```
srowen, line 64: This re-declares the array every time the method is called, and reinstantiates all the objects. It isn't necessary, which is why the code is the way it is now.

core/src/main/java/com/google/zxing/client/result/ResultParser.java
```diff
@@ -152,9 +148,15 @@ protected static int parseHexDigit(char c) {
   }
 
   protected static boolean isStringOfDigits(CharSequence value, int length) {
-    return value != null && length > 0 && length == value.length() && DIGITS.matcher(value).matches();
+    boolean isDigitsMatch = value != null && length > 0 && length == value.length() && isDigits(value);
```
srowen, line 151: I don't think the extra variable adds anything here

core/src/main/java/com/google/zxing/client/result/ResultParser.java
```diff
@@ -163,26 +165,27 @@ protected static boolean isSubstringOfDigits(CharSequence value, int offset, int
...
   }
 
-  static Map<String,String> parseNameValuePairs(String uri) {
+  static String[] parseNameValuePairs(String uri) {
     int paramStart = uri.indexOf('?');
     if (paramStart < 0) {
       return null;
     }
-    Map<String,String> result = new HashMap<>(3);
+    String[] result = new String[3];
```
srowen, line 173: This does not work. 3 is an expected size to the hashmap, but is a fixed size for an array. This will keep overwriting the same value

core/src/main/java/com/google/zxing/client/result/ResultParser.java
```diff
@@ -221,24 +224,27 @@ static String[] matchPrefixedField(String prefix, String rawText, char endChar,
...
           more = false;
         }
       }
     }
-    if (matches == null || matches.isEmpty()) {
+    if (matches == null || matches.length() == 0) {
       return null;
     }
-    return matches.toArray(EMPTY_STR_ARRAY);
+    return matches.toString().split("\n");
```
srowen, line 247: This is less efficient, no

core/src/main/java/com/google/zxing/client/result/VINResultParser.java
```diff
@@ -205,5 +180,90 @@ private static String countryCode(CharSequence wmi) {
     }
     return null;
   }
+  
+  private static int getModelYear(char c) {
```
srowen, line 184: This is overkill design-wise and adds runtime overhead

srowen (commented):
OK thanks for the effort, but I think these changes are not improvements

PR #1747, opened 2024-01-17
Support for platforms where SJIS is unavailable

blakepuhak:
Certain embedded JREs do not include lib/charsets.jar  This means that these platforms don't support Charsets outside of the six [standard Charsets](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/Charset.html).  

#1330 made the assumption that the Shift JIS Charset *would* be available, so on platforms without Shift JIS support, attempting to load QRCodeWriter results in an UnsupportedCharsetException before attempting to encode any text.

This change, like the change in commit [a6273e3](https://github.com/zxing/zxing/commit/a6273e3bc788135e4ad755fe6a7c1a4bf74fcdc9), handles UnsupportedCharsetExceptions in the static initializers, then checks for the properly loaded SJIS Charset during encode and decode operations.

So as to not introduce any new test dependencies, the *tests* still assume that the Shirt JIS and EUC_JP Charsets will be available.

### DIFF
--- a/core/src/main/java/com/google/zxing/common/StringUtils.java
+++ b/core/src/main/java/com/google/zxing/common/StringUtils.java
@@ -54,10 +54,20 @@ public final class StringUtils {
     }
     GB2312_CHARSET = gb2312Charset;
   }
-  private static final Charset EUC_JP = Charset.forName("EUC_JP");
+  private static final Charset EUC_JP;
+  static {
+    Charset eucJpCharset;
+    try {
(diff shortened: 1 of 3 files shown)


srowen (approved):
I think that's reasonable, though in such situations you are maybe going to get different (wrong) decode results. Maybe better than none at all

blakepuhak:
My use case is for encoding, but I didn't want to commit half a fix and leave someone else with a potential NullPointerException while decoding on a platform where SJIS isn't available.

PR #1780, opened 2024-03-31
University assingment refactoring technique.

vyansidiyora2000:


srowen:
Thanks, but I don't know if this is an improvement in all cases, and not worth changing this old code

PR #1784, opened 2024-04-08
I have added comments to explain the purpose of each calculation, imp…

rajveerisme:
…roving code readability.

I've retained the original logic for calculating newLeft, newTop, newRight, and newBottom as it seems efficient and appropriate for its purpose.


srowen:
Thank you, but I don't think these comments add much of anything

rajveerisme:
Thank you for your feedback sir , i will continue working on more corrections

PR #1785, opened 2024-04-09
avoid using additional streams like 'ByteArrayOutputStream' and 'ByteArrayInputStream'

rajveerisme:
I am still using the same tempFile variable to create a temporary file.
The image data is written to this temporary file.
After reading the image data, we delete the temporary file within the finally block to ensure proper cleanup even in case of exceptions.


srowen (commented):
This PR includes unrelated changes, and I don't believe improves any behavior of temp files. Temp files were already deleted.

PR #1801, opened 2024-04-30
Promote renderResult method to public

cdcarloschacon:
This small commit makes public the "`renderResult`" method of the "`QRCodeWriter`" class.

The reason to promote this method as public is that I'm facing business case where, besides of generating the QR code as a PNG image, I also need to print information about the QR code itself (specifically, the QR version). Navigating through the code of zxing library, I noticed that the `com.google.zxing.qrcode.encoder.Encoder` allows me to have that (the QR metadata); however, by invoking that class I only have part of the equation (the `QRCode` DTO) ... I'm still needing to generate the `BitMatrix` in order to produce the QR image.

Here, I have two alternatives: either I invoke the `Encoder.encode` class by myself and repeat the same logic when invoking the `QRCodeWriter.encode` method; or I make the `renderResult` Public, call the `Encoder.encode` and then I proceed to call the `QRCodeWriter.renderResult` method with the previous result.

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/QRCodeWriter.java
+++ b/core/src/main/java/com/google/zxing/qrcode/QRCodeWriter.java
@@ -79,9 +79,27 @@ public final class QRCodeWriter implements Writer {
     return renderResult(code, width, height, quietZone);
   }
 
-  // Note that the input matrix uses 0 == white, 1 == black, while the output matrix uses
-  // 0 == black, 255 == white (i.e. an 8 bit greyscale bitmap).
-  private static BitMatrix renderResult(QRCode code, int width, int height, int quietZone) {
+  /**
+   * Renders the given {@link QRCode} as a {@link BitMatrix}, scaling the
+   * same to be compliant with the provided dimensions.
+   *
+   * <p>If no scaling is required, both {@code width} and {@code height}
+   * arguments should be non-positive numbers.
+   *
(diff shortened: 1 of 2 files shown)


cdcarloschacon:
@srowen , the javadoc was added to the target method and some junit tests were added too.

core/src/test/java/com/google/zxing/qrcode/QRCodeWriterTestCase.java
```diff
@@ -133,4 +136,55 @@ public void testRegressionTest() throws Exception {
         "renderer-test-01.png");
   }
 
+  @Test
+  public void renderResultScalesNothing() {
+    final int expected_size = 33;           // Original Size (25) + quietZone
```
srowen, line 141: Oh, this fails style checks. Just `expectedSize`

core/src/main/java/com/google/zxing/qrcode/QRCodeWriter.java
```diff
@@ -37,6 +37,60 @@ public final class QRCodeWriter implements Writer {
 
   private static final int QUIET_ZONE_SIZE = 4;
 
+  /**
```
srowen, line 40: I'm not against moving this, but it'd be slightly cleaner to not move the method

srowen (changes requested):

cdcarloschacon:
I will fix that ASAP ... thanks for the insights ...

PR #1849, opened 2024-09-04
Update CameraConfigurationUtils.java

MarlonHerrera01:


android-core/src/main/java/com/google/zxing/client/android/camera/CameraConfigurationUtils.java
```diff
@@ -299,18 +299,17 @@ public static Point findBestPreviewSizeValue(Camera.Parameters parameters, Point
       int realWidth = size.width;
       int realHeight = size.height;
       int resolution = realWidth * realHeight;
-      if (resolution < MIN_PREVIEW_PIXELS) {
+      double distortion = Math.abs(aspectRatio - screenAspectRatio);
```
srowen, line 302: You haven't explained the change.
This does not compile as aspectRatio is defined below.
I don't think fixing that makes this a worthwhile change though

PR #1870, opened 2024-10-08
Be more stringent on email validation rules

### LINKED ISSUE #1818: Any barcode with an @ gets misidentified as an email address, disabling custom/web search
Steps to reproduce (1):

Create a barcode for XY@asdf
Scan the barcode

Result: the barcode is correctly scanned and decoded, with two buttons at the bottom of the screen: "Send email" and "Add contact".

Expected result: same, but with buttons for "Web search" and "Custom search" (if set up).

Test case 2:
Same, but with barcode for @asdf

While I don't mind that the application considers anything which has an "@" inside worthy to be offered the option for email-related functions, it prevents easy searching for email addresses on the web and also scanning of product codes with an "@". And no, I have no clue who had this funny idea, but in this case, the first part before the "@" is a company code and the second part is the item code (yes, *that* order exactly...).

Proposed fix:

In EmailAddressResultHandler.java, lines 32-33 are
```
    R.string.button_email,
    R.string.button_add_contact
```
and might be changed to:
```
    R.string.button_custom_product_search,
    R.string.button_web_search,
    R.string.button_email,
    R.string.button_add_contact
```

For further consideration: I am undecided whether it would be a good idea to have custom search and web search buttons for every scan result. I guess it doesn't make sense for a VCard (lots of data inside) but might be nice for telephone numbers.

Anyone, any opinion?

ftiercelin:
- [x] be more stringent on email validation rules
- [x] add UT to validate various email formats

Fixes #1818

### DIFF
--- a/core/src/main/java/com/google/zxing/client/result/EmailDoCoMoResultParser.java
+++ b/core/src/main/java/com/google/zxing/client/result/EmailDoCoMoResultParser.java
@@ -29,8 +29,8 @@ import java.util.regex.Pattern;
  */
 public final class EmailDoCoMoResultParser extends AbstractDoCoMoResultParser {
 
-  private static final Pattern ATEXT_ALPHANUMERIC = Pattern.compile("[a-zA-Z0-9@.!#$%&'*+\\-/=?^_`{|}~]+");
-
+  private static final Pattern EMAIL = Pattern.compile("^[^:]+@([0-9a-zA-Z\\\\-]+\\.)+[0-9a-zA-Z\\\\-]{2,}$");
+  
   @Override
   public EmailAddressParsedResult parse(Result result) {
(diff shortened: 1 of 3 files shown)


srowen:
Looks pretty reasonable, just need to appease the linter:

https://github.com/zxing/zxing/actions/runs/11241558163/job/31253750013?pr=1870#step:4:1252

srowen:
Hm, still seems like there are some test failures here

ftiercelin:
yes indeed, this test is failing:
```java
    doTestResult("srowen@example", "srowen@example", ParsedResultType.EMAIL_ADDRESS);
```

and rightly so, as this shouldn't be an email address. I'll change the UT and add a new one

PR #1898, opened 2025-01-22
Implemented a tryHarder for the WhiteRectangeDetector

pmneo:
Proposal for a try harder variant of the WhiteRectableDetector to find a not centered Aztec and Data Matrix Code in a page


srowen:
(The code format would need a fix)
It's reasonable, though I'm concerned that it may spend a ton of CPU cycles with unclear additional benefit. I'm not clear +50 is the right stride, especially for very large images.
Describe more what the idea is here - when the code is not in the center? then what about looking in the four quadrants? that kind of thing

pmneo:
Here is a sample image of the not found barcode:

undetected_barcode

With this image the detector failes without my adjustments.

Issue #1900, opened 2025-01-30
Wrong dimension


magethle:
Hello,

I have a qr code with version 29 which zxing is not able to parse.
It fails because zxing computes a dimension of 135 (in this [line](https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/qrcode/detector/Detector.java#L205)) which does not fulfil 4*version +17 (=133). If the difference is only 1 zxing corrects the computed dimension (e.g. for 134 it would return 133, for 136 it would return 137). Why is an error of 2 not accepted?

If I change [this line](https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/qrcode/detector/Detector.java#L214) to `dimension=dimension-2;` so that it returns 133 everything works and zxing can read the qr code.

Here is the qr code. It's the 1st part of a structured append qr code and contains part of a base64 string:
![Image](https://github.com/user-attachments/assets/ee89052f-eb71-464c-897d-52b865cecff4)

srowen:
135 is ambiguous; should you fix it to 133 or 137? the primary use case is decoding from a camera where you can just try again on the next image.

This image should be decode with PURE_BARCODE mode, which should work.

I can see the argument for fixing the dimension in this case; fixing 135 to 133 might be wrong but otherwise you're just failing. Avoiding false positives is important, so you want to fail if decoding is looking bad, but in the end QR codes have checksums so it'd be very hard to get a false positive.

You are welcome to try a change to that effect and see what it does to tests.

PR #1901, opened 2025-02-03
Add hint to correct wrong dimension

magethle:
Adds a DecodeHintType to correct a wrong dimension to upper or lower  value.

If dimension is wrong only by 1 it is already corrected automatically. With this proposal also errors by 2 can be corrected (to lower or upper value).

Old code remains unaffected as the new behavior has to be activated explicitly.

Fixes issue #1900

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/detector/Detector.java
+++ b/core/src/main/java/com/google/zxing/qrcode/detector/Detector.java
@@ -211,7 +211,7 @@ public class Detector {
         dimension--;
         break;
       case 3:
-        throw NotFoundException.getNotFoundInstance();
+        dimension = dimension - 2;
     }
     return dimension;
   }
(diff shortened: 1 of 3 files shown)


srowen:
I wouldn't have one hint for it, let alone two. Just pick one direction and always round it. I'd just round down

magethle:
> I wouldn't have one hint for it, let alone two. Just pick one direction and always round it. I'd just round down

Ok, I always round down now.

core/src/main/java/com/google/zxing/qrcode/detector/Detector.java
```diff
@@ -211,7 +211,7 @@ private static int computeDimension(ResultPoint topLeft,
         dimension--;
         break;
       case 3:
-        throw NotFoundException.getNotFoundInstance();
+        dimension = dimension - 2;
```
srowen, line 214: Total nit, but can you write:
```
  dimension -= 2;
  break;
```
I think the logic is OK.

PR #1908, opened 2025-02-07
Fix distortion calculation for isosceles right triangles

magethle:
Fix distortion calculation for isosceles right triangles as discussed in https://github.com/zxing/zxing/discussions/1906

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
+++ b/core/src/main/java/com/google/zxing/qrcode/detector/FinderPatternFinder.java
@@ -695,6 +695,8 @@ public class FinderPatternFinder {
           // we need to check both two equal sides separately.
           // The value of |c^2 - 2 * b^2| + |c^2 - 2 * a^2| increases as dissimilarity
           // from isosceles right triangle.
+          // Heuristically it seems that the following formula works better (although it's
+          // not clear any more why...)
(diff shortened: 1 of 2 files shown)


srowen:
Nice, thank you. The result is virtually identical, except that actually 1 fewer image passes tests.

I could go either way on this. Either accept that this change is more principled, and change the test to allow the additional failure, or, decide to leave it alone as a likely mistake that happens to work OK (or just change the comment). Do you have a preference?

magethle:
Well, I don't have a strong preference either.
Probably it's better (more correct) to change the code than the comment.
Should I change the test then?

srowen:
Yeah I'm fine with that.

magethle:
Actually it's more than just 1 image.
It fails no to detect the following qrcodes in qrcode-3:
rotation of 180°: 10.png, 11.png and 13.png
rotation of 270°: 13.png

srowen:
Hm, I'm conflicted here. It seems like it only reduces detection a little bit. There are no tests where more are decoded (you would see "too lax..." messages in the logs). While the change makes sense, maybe the current code makes more sense in practice. We could leave it alone, or, update the comment, but I'm not sure how to explain this heuristic.

magethle:
I reverted the change and added a comment but was unsure how to explain it.
You can merge this PR or close it unchanged.

PR #2014, opened 2025-11-11
Loosen ITF max variance for 3x module widths

gredler:
ITF can use 2x or 3x wide bars, and it is natural for the 3x bars to have larger variances than the 2x wide bars. This change takes this into account, so that fewer 3x ITF symbols are rejected.


gredler:
Awesome, thanks!

PR #2039, opened 2026-01-14
Update ECIEncoderSet.java

grandiobit:
Now encoding hints work as it's intented


grandiobit:
In most cases this will lead to situation when encoding hint doesnt' work for encoders in the bottom of the list (like Windows encoding pages), because there was encoding that fits before it. In some use cases it can be crucial. Maybe there is a better solution for this problem (list sorting based on encoding hint for example), my solution may be not so efficient

Issue #1567, opened 2022-11-01
The library is hanging when processing a triggering image


hauck-jvsh:
As said before in #1541, we integrate this library in another project (https://github.com/sepinf-inc/IPED/) to detect images with QRCodes. Another image with approximately 2Mb triggers an unexpected behavior, the library is running for hours without finishing.  The triggering file is attached. The thread dump below shows that the process is locked inside the zxing library.

>  java.lang.Thread.State: RUNNABLE
	at com.google.zxing.qrcode.detector.FinderPatternFinder.selectBestPatterns(FinderPatternFinder.java:637)
	at com.google.zxing.qrcode.detector.FinderPatternFinder.find(FinderPatternFinder.java:164)
	at com.google.zxing.qrcode.detector.Detector.detect(Detector.java:81)
	at com.google.zxing.qrcode.QRCodeReader.decode(QRCodeReader.java:77)
	at com.google.zxing.qrcode.QRCodeReader.decode(QRCodeReader.java:64)
	at iped.engine.task.QRCodeTask.process(QRCodeTask.java:143)

I take a small look in the code and seems to be a method with a cubic complexity in relation to the number of possible centers. Maybe this number should have an upper bound?

[triggeringImage.zip](https://github.com/zxing/zxing/files/9913238/triggeringImage.zip)

srowen:
Yeah you have it right. I think there is an easy fix. Coming shortly. Are you in a position to build the library from source and try the fix?

lfcnassif:
Just confirmed the fix, thank you @srowen!

lfcnassif:
Out of curiosity, is there a time frame for 3.5.1?

srowen:
Just when I have a reason to make a release. It's been 6 months and some reasonable fixes since the last one so I could do it in the next couple days.

hauck-jvsh:
Wow, you were fast I was in a meeting and when I came out the problem was already fixed and tested.

srowen:
3.5.1 should be released now. I needed to get out some other fixes, so good timing

lfcnassif:
Great, thank you!

PR #2061, opened 2026-04-07
Fix QR decode failure for small/native-size images

### LINKED ISSUE #1976: Unable to parse a QR code that was generated from ZXing itself after 110ef9e
In response to #1567, ZXing added 110ef9e. However, this has caused ZXing to fail to decode images that it itself has created. I have attached a minimal repro below. The problem may lie in `FinderPatternFinder::find`, as it seems to believe that two of the three finder patterns in the image only have a `count` of 1.

```java
package com.google.zxing.client.j2se;

import com.google.zxing.BarcodeFormat;
import com.google.zxing.BinaryBitmap;
import com.google.zxing.MultiFormatReader;
import com.google.zxing.common.BitMatrix;
import com.google.zxing.common.HybridBinarizer;
import com.google.zxing.qrcode.QRCodeWriter;
import org.junit.Assert;
import org.junit.Test;

import java.awt.image.BufferedImage;
import java.util.UUID;

public class QrCodeRoundTripTestCase extends Assert {
  @Test
  public void testRoundTrip() throws Exception {
    String payload = UUID.randomUUID().toString();

    QRCodeWriter writer = new QRCodeWriter();
    BitMatrix matrix = writer.encode(payload, BarcodeFormat.QR_CODE, 0, 0);
    BufferedImage image = MatrixToImageWriter.toBufferedImage(matrix);

    BufferedImageLuminanceSource source = new BufferedImageLuminanceSource(image);
    BinaryBitmap bitmap = new BinaryBitmap(new HybridBinarizer(source));
    MultiFormatReader reader = new MultiFormatReader();

    assertEquals(reader.decode(bitmap).getText(), payload);
  }
}
```

jabagawee:
## Summary
- Fix #1976: QR codes at native/small sizes fail to decode because the
  CENTER_QUORUM filter (110ef9e, fixing #1567) removes legitimate finder
  patterns that were only detected once due to the row-skip interval.
- Make the filter conditional: only apply when >= 3 confirmed patterns
  exist or when the candidate list is large enough that O(n^3) cost
  matters.
- Add regression tests for both #1976 (round-trip encode/decode at
  native size) and #1567 (noisy image completes within 5 seconds).

## Root cause

The `find()` method scans rows with `iSkip=3` (MIN_SKIP) for small
images. For a 33x33 native-size QR (version 2, 25 modules + 4-module
quiet zone), the finder pattern centers span ~3 rows at 1px/module.
With iSkip=3, only one scan row crosses each top finder pattern center.
After the first detection, iSkip drops to 2, so the bottom-left pattern
gets a second hit. Result: top-left count=1, top-right count=1,
bottom-left count=2. The CENTER_QUORUM filter (count >= 2) removes the
top patterns, leaving 1 of 3 needed.

## Why this approach

The #1567 fix addressed a real performance problem: noisy images with
no QR code can produce thousands of spurious finder pattern candidates,
and the O(n^3) triangle-selection loop hangs. Simply removing the
filter re-opens that case.

The fix bounds the unfiltered path's cost with a candidate-list size
threshold. Below the threshold (25 candidates, ~2300 loop iterations),
the cubic analysis is negligible, so filtering is unnecessary. Above
it, the filter applies unconditionally. A clean single-QR image
produces ~3 candidates; even multi-code images stay well under 25.

Alternatives considered and rejected:
- Remove filter entirely: passes all tests but re-opens #1567 for
  pathological inputs.
- Reduce MIN_SKIP: treats symptom, not cause; changes scan behavior
  globally for small images.
- Try-with-filter then retry-without: re-opens #1567 when fewer
  than 3 confirmed patterns exist alongside many unconfirmed ones.

## Test plan
- [x] Existing core tests pass (562 tests including new ones)
- [x] New round-trip test: encode with width=0, height=0, decode
  succeeds
- [x] New performance test: original #1567 triggering image completes
  within 5 seconds (times out without the size threshold)

This contribution is my original work and I license it to the project
under the project's open source license.


srowen (changes requested):
I think the 5 second timeout isn't enough, not for Github runners at least. 30 seconds?

jabagawee:
my bad, i made some faulty assumptions about how bad it would be in the worst case. updated to 30 seconds.

srowen (approved):

PR #2066, opened 2026-04-28
Add Aztec reader initialization support

gredler:
This PR adds support for decoding "reader initialization" Aztec Code symbols. Without this change, these symbols result in a FormatException, because during decoding it looks like the layer count and data count aren't compatible.


gredler:
I plan to add support for reader initialization in Code 128, Data Matrix and PDF 417 in separate individual PRs, but I can incorporate all four into this PR if that's easier. Just let me know.

srowen (approved):

PR #2069, opened 2026-04-30
Add reader initialization support for Data Matrix and PDF417

gredler:
This PR adds support for reader initialization to both Data Matrix and PDF417. They both needed a change to `DecoderResult`, so I decided to include them both here.

Per the Data Matrix spec, Data Matrix symbols are interpreted as reader initialization symbols if the first codeword in the symbol is the reader programming character (234).

Per the PDF417 spec, PDF417 symbols are interpreted as reader initialization symbols if codeword 921 appears immediately following the symbol length descriptor (which is the first codeword).

I think these are the last symbols which need to support reader initialization.


srowen (approved):

srowen:
Thanks! nice and clean work in these PRs.

PR #2077, opened 2026-05-27
Avoid charset provider lookups for UTF-8 QR encoding

### LINKED ISSUE #2076: QR encoding can trigger Android StrictMode disk reads via charset lookup
### Summary

Encoding a QR code with ZXing `core` can trigger Android `StrictMode` disk-read violations during charset / ECI initialization. This is not a QR correctness problem, but it is surprising for Android callers because a normal in-process QR encode can cause lazy classpath / APK zip access.

I searched existing issues for `StrictMode`, `DiskReadViolation`, `ServiceLoader`, `Charset.isSupported`, `CharacterSetECI`, and `ZipFile`. I found related ECI / charset issues such as #1444 and #1042, but did not find an issue about Android StrictMode disk I/O from the charset lookup path.

### Version / environment

- ZXing core: 3.5.4
- Platform: Android 16, Pixel 9
- App code path: `MultiFormatWriter().encode(..., BarcodeFormat.QR_CODE, ..., hints)`
- The app has debug `StrictMode.ThreadPolicy.detectAll()` enabled.

### Repro shape

A minimal Android repro is roughly:

```java
StrictMode.setThreadPolicy(new StrictMode.ThreadPolicy.Builder()
    .detectDiskReads()
    .penaltyLog()
    .build());

Map<EncodeHintType,Object> hints = new EnumMap<>(EncodeHintType.class);
hints.put(EncodeHintType.CHARACTER_SET, StandardCharsets.UTF_8.name());
new MultiFormatWriter().encode(
    "WIFI:S:\u6d4b\u8bd5;T:WPA;P:password;;",
    BarcodeFormat.QR_CODE,
    512,
    512,
    hints);
```

Observed in Crashlytics / StrictMode:

```text
android.os.strictmode.DiskReadViolation
  at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1836)
  at java.util.zip.ZipFile$Source.initCEN(ZipFile.java:1860)
  at java.util.jar.JarFile.<init>(JarFile.java:183)
  at libcore.io.ClassPathURLStreamHandler.<init>(ClassPathURLStreamHandler.java:52)
  at dalvik.system.DexPathList$Element.findResource(DexPathList.java:780)
  at dalvik.system.BaseDexClassLoader.findResources(BaseDexClassLoader.java:330)
  at java.lang.ClassLoader.getResources(ClassLoader.java:1455)
  at java.util.ServiceLoader$LazyClassPathLookupIterator.nextProviderClass(ServiceLoader.java:1157)
  ...
```

The exact StrictMode frame varies between `read`, `lseek`, `FileChannelImpl.map`, etc., but all samples point at APK / zip metadata being read as part of classpath resource lookup.

### Why this seems to come from ZXing

The QR encoder path touches Java charset APIs during encode. In `core` 3.5.4, the relevant paths appear to be:

- `com.google.zxing.qrcode.encoder.Encoder.encode(...)` calls `Charset.forName(...)` when `EncodeHintType.CHARACTER_SET` is supplied.
- `com.google.zxing.common.CharacterSetECI` initializes supported ECI mappings with `Charset.isSupported(...)`.
- `com.google.zxing.common.StringUtils` initializes Shift-JIS / GB2312 / EUC-JP charset support with `Charset.forName(...)`.

On Android, these charset calls can go through provider discovery, which uses `ServiceLoader` and `ClassLoader.getResources(...)`. That can inspect APK zip entries and trip StrictMode disk-read detection.

### Expected behavior

For common QR encoding cases, especially with known standard charsets such as ISO-8859-1 / UTF-8 / Shift_JIS, ZXing should ideally avoid triggering Android classpath resource scanning during encode, or at least defer the broad charset-provider probing until a caller actually needs a non-built-in charset.

Possible directions, if this is considered in scope:

- Prefer `StandardCharsets` / known charset constants for common ECI mappings where possible.
- Avoid populating all `CharacterSetECI` support via `Charset.isSupported(...)` during class initialization on Android-sensitive paths.
- Lazily resolve optional charsets only when the caller requests them.

I realize the project is in maintenance mode and bug reports without PRs are generally less actionable. Filing this first because the behavior is Android-specific and the right compatibility/performance tradeoff is probably worth maintainer input before a narrow PR.

Mygod:
Fixes #2076.

## Summary

This avoids unnecessary charset-provider lookups when encoding QR codes with standard character set hints such as UTF-8.

The UTF-8 QR path can currently initialize charset lookup machinery that probes optional charset providers. On Android this can perform disk I/O, which is visible to StrictMode when QR codes are encoded on the main thread.

This change:

- Lazily initializes `CharacterSetECI` lookup maps instead of probing `Charset` support during enum class initialization.
- Uses `StandardCharsets` for guaranteed standard ECI charsets, including UTF-8.
- Keeps ECI support checks tied to the primary ECI charset, so aliases do not make an unsupported ECI appear decodable.
- Avoids touching `StringUtils.SHIFT_JIS_CHARSET` during QR mode selection unless the requested charset is Shift_JIS-like.

## Validation

- `mvn -pl core -Dtest=CharacterSetECITestCase,EncoderTestCase,DecodedBitStreamParserTestCase test`
- `mvn -pl core test`
- `git diff --check`

### DIFF
--- a/core/src/main/java/com/google/zxing/common/CharacterSetECI.java
+++ b/core/src/main/java/com/google/zxing/common/CharacterSetECI.java
@@ -20,7 +20,6 @@ import com.google.zxing.FormatException;
 
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.nio.charset.UnsupportedCharsetException;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -34,7 +33,7 @@ public enum CharacterSetECI {
 
   // Enum name is a Java encoding valid for java.lang and java.io
   Cp437(new int[]{0,2}),
-  ISO8859_1(new int[]{1,3}, "ISO-8859-1"),
+  ISO8859_1(new int[]{1,3}, StandardCharsets.ISO_8859_1, "ISO-8859-1"),
   ISO8859_2(4, "ISO-8859-2"),
   ISO8859_3(5, "ISO-8859-3"),
   ISO8859_4(6, "ISO-8859-4"),
@@ -54,30 +53,29 @@ public enum CharacterSetECI {
   Cp1251(22, "windows-1251"),
   Cp1252(23, "windows-1252"),
   Cp1256(24, "windows-1256"),
-  UnicodeBigUnmarked(25, "UTF-16BE", "UnicodeBig"),
-  UTF8(26, "UTF-8"),
-  ASCII(new int[] {27, 170}, "US-ASCII"),
+  UnicodeBigUnmarked(25, StandardCharsets.UTF_16BE, "UTF-16BE", "UnicodeBig"),
+  UTF8(26, StandardCharsets.UTF_8, "UTF-8"),
+  ASCII(new int[] {27, 170}, StandardCharsets.US_ASCII, "US-ASCII"),
   Big5(28),
   GB18030(29, "GB2312", "EUC_CN", "GBK"),
   EUC_KR(30, "EUC-KR");
 
-  private static final class Maps {
-    private static final Map<Integer,CharacterSetECI> VALUE_TO_ECI = new HashMap<>();
-    private static final Map<String,CharacterSetECI> NAME_TO_ECI = new HashMap<>();
-    static {
-      for (CharacterSetECI eci : values()) {
-        for (int value : eci.values) {
-          VALUE_TO_ECI.put(value, eci);
-        }
-        NAME_TO_ECI.put(eci.name(), eci);
-        for (String name : eci.otherEncodingNames) {
-          NAME_TO_ECI.put(name, eci);
-        }
+  private static final Map<Integer,CharacterSetECI> VALUE_TO_ECI = new HashMap<>();
+  private static final Map<String,CharacterSetECI> NAME_TO_ECI = new HashMap<>();
+  static {
+    for (CharacterSetECI eci : values()) {
+      for (int value : eci.values) {
+        VALUE_TO_ECI.put(value, eci);
+      }
+      NAME_TO_ECI.put(eci.name(), eci);
+      for (String name : eci.otherEncodingNames) {
+        NAME_TO_ECI.put(name, eci);
       }
     }
   }
 
   private final int[] values;
+  private final Charset standardCharset;
   private final String[] otherEncodingNames;
 
   CharacterSetECI(int value) {
@@ -85,12 +83,20 @@ public enum CharacterSetECI {
   }
 
   CharacterSetECI(int value, String... otherEncodingNames) {
-    this.values = new int[] {value};
-    this.otherEncodingNames = otherEncodingNames;
+    this(new int[] {value}, null, otherEncodingNames);
+  }
+
+  CharacterSetECI(int value, Charset standardCharset, String... otherEncodingNames) {
+    this(new int[] {value}, standardCharset, otherEncodingNames);
   }
 
   CharacterSetECI(int[] values, String... otherEncodingNames) {
+    this(values, null, otherEncodingNames);
+  }
+
+  CharacterSetECI(int[] values, Charset standardCharset, String... otherEncodingNames) {
     this.values = values;
+    this.standardCharset = standardCharset;
     this.otherEncodingNames = otherEncodingNames;
   }
 
@@ -99,34 +105,14 @@ public enum CharacterSetECI {
   }
 
   public Charset getCharset() {
-    Charset charset = getPrimaryCharsetOrNull();
-    if (charset != null) {
-      return charset;
+    if (standardCharset != null) {
+      return standardCharset;
     }
-    throw new UnsupportedCharsetException(name());
-  }
-
-  private Charset getPrimaryCharsetOrNull() {
-    switch (this) {
-      case ISO8859_1:
-        return StandardCharsets.ISO_8859_1;
-      case UTF8:
-        return StandardCharsets.UTF_8;
-      case ASCII:
-        return StandardCharsets.US_ASCII;
-      case UnicodeBigUnmarked:
-        return StandardCharsets.UTF_16BE;
-      default:
-    }
-    try {
-      return Charset.forName(name());
-    } catch (UnsupportedCharsetException uce) {
-      return null;
-    }
+    return Charset.forName(name());
   }
 
   private boolean isCharsetSupported() {
-    return getPrimaryCharsetOrNull() != null;
+    return standardCharset != null || Charset.isSupported(name());
   }
 
   /**
@@ -135,7 +121,7 @@ public enum CharacterSetECI {
    *   but unsupported
    */
   public static CharacterSetECI getCharacterSetECI(Charset charset) {
-    CharacterSetECI eci = Maps.NAME_TO_ECI.get(charset.name());
+    CharacterSetECI eci = NAME_TO_ECI.get(charset.name());
     return eci == null || eci.isCharsetSupported() ? eci : null;
   }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/google/zxing/common/CharacterSetECI.java
```diff
@@ -99,7 +99,34 @@ public int getValue() {
...
   public Charset getCharset() {
-    return Charset.forName(name());
+    Charset charset = getPrimaryCharsetOrNull();
+    if (charset != null) {
+      return charset;
+    }
+    throw new UnsupportedCharsetException(name());
+  }
+
+  private Charset getPrimaryCharsetOrNull() {
```
srowen: This feels complex. Maybe the enum class can have a field like "standardCharset" which is null by default, but can be given a reference to a StandardCharsets object in the constructor if applicable. Then return that here instead of calling forName(), if it exists.

core/src/main/java/com/google/zxing/common/CharacterSetECI.java
```diff
@@ -60,12 +61,11 @@ public enum CharacterSetECI {
   GB18030(29, "GB2312", "EUC_CN", "GBK"),
   EUC_KR(30, "EUC-KR");
 
-  // only character sets supported by the current JVM are registered here
-  private static final Map<Integer,CharacterSetECI> VALUE_TO_ECI = new HashMap<>();
-  private static final Map<String,CharacterSetECI> NAME_TO_ECI = new HashMap<>();
-  static {
-    for (CharacterSetECI eci : values()) {
-      if (Charset.isSupported(eci.name())) {
+  private static final class Maps {
```
srowen: Is this an attempt to lazy-load?

core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
```diff
@@ -276,6 +277,11 @@ private static Mode chooseMode(String content, Charset encoding) {
     return Mode.BYTE;
   }
 
+  private static boolean isShiftJIS(Charset encoding) {
```
srowen: This seems redundant, is it also just an attempt to not load the class?

core/src/main/java/com/google/zxing/common/CharacterSetECI.java
```diff
@@ -108,7 +135,8 @@ public Charset getCharset() {
    *   but unsupported
    */
   public static CharacterSetECI getCharacterSetECI(Charset charset) {
-    return NAME_TO_ECI.get(charset.name());
+    CharacterSetECI eci = Maps.NAME_TO_ECI.get(charset.name());
+    return eci == null || eci.isCharsetSupported() ? eci : null;
```
srowen, line 125: The map can just contain null values for unsupported charsets, I suppose. It doesn't have to be handled here I think

srowen (commented):
If this can be a bit simpler I think it could be mergeable.
Why do you need this strict mode with a barcode app?

core/src/main/java/com/google/zxing/common/CharacterSetECI.java
Mygod: Done in c1ff33453. I added a standardCharset field for the ECIs backed by StandardCharsets, so getCharset() returns that directly before falling back to Charset.forName(name()).

core/src/main/java/com/google/zxing/common/CharacterSetECI.java
Mygod: Yes, that was the intent of the inner Maps holder: avoid eager charset-provider probing during CharacterSetECI class initialization. I simplified this in c1ff33453 by building the maps normally and moving the supported-runtime check to the point where an ECI is actually requested.

core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
Mygod: Removed in c1ff33453. The Shift_JIS check is now inline with the existing Kanji-mode path, and it still only uses Kanji mode when the resolved charset equals StringUtils.SHIFT_JIS_CHARSET.

core/src/main/java/com/google/zxing/common/CharacterSetECI.java
Mygod: I kept this as a lookup-time support check rather than putting nulls in the maps. To know which map entries should be null during static initialization, this would still need eager Charset.isSupported(name()) calls for every ECI, which is the provider path this PR is avoiding. The current version builds the maps without provider checks and only probes support for the requested non-standard ECI.

Mygod:
Pushed c1ff33453 to simplify the follow-up.

On StrictMode: this is for ZXing used as a library in Android apps that enable StrictMode, not for the ZXing app specifically. Encoding a UTF-8 QR code with CHARACTER_SET can currently reach charset-provider lookup on the caller thread, and Android can report that as disk-read I/O. The change uses StandardCharsets for the required standard charsets so common ECI lookups avoid that provider path, while optional/non-standard charsets keep the existing runtime-support behavior.

Validation:
- mvn -pl core -Dtest=CharacterSetECITestCase,EncoderTestCase,DecodedBitStreamParserTestCase test
- mvn -pl core test

core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
```diff
@@ -249,7 +249,9 @@ public static Mode chooseMode(String content) {
    * if it is Shift_JIS, and the input is only double-byte Kanji, then we return {@link Mode#KANJI}.
    */
   private static Mode chooseMode(String content, Charset encoding) {
-    if (StringUtils.SHIFT_JIS_CHARSET != null &&
+    if (encoding != null &&
+        ("Shift_JIS".equalsIgnoreCase(encoding.name()) || "SJIS".equalsIgnoreCase(encoding.name())) &&
```
srowen, line 253: Last question, what is the need for this? I don't think it avoids loading StringUtils

srowen (commented):
I think the change is fine, one last question though

core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java
Mygod: It avoids initializing StringUtils for non-Shift_JIS encodings. The costly part is not the import/reference itself, but accessing StringUtils.SHIFT_JIS_CHARSET (a static final Charset), which runs StringUtils static initialization and probes several optional charsets. For UTF-8 and other byte encodings, encoding.name() lets this path short-circuit before that static field access. Shift_JIS still initializes StringUtils, because the existing Kanji encoder path depends on the resolved SHIFT_JIS_CHARSET.

srowen: Isn't the static init going to run the moment you import the class at the top? I thought you'd have to use reflection to achieve what you're talking about

Mygod: No, imports are compile-time only; they do not initialize the imported class. Reflection is not needed for this. The runtime trigger here would be the getstatic for StringUtils.SHIFT_JIS_CHARSET, because it is a Charset field rather than a compile-time constant. The preceding encoding.name() check short-circuits before that field access for UTF-8 and other non-Shift_JIS encodings. I also checked the bytecode: the StringUtils.SHIFT_JIS_CHARSET getstatic is emitted after the name checks.

srowen: Got it. I think I've forgotten so much Java.

Mygod: No problem. Nowadays I delegate all these pesky details to the coding agent. ;)

PR #2085, opened 2026-06-04
avoid out-of-bounds reads after control codewords in pdf417 parser

jmestwa-coder:
Several spots in the pdf417 parser read the codeword right after a control codeword without checking another one is there:
- decode: the byte value after a 913 mode shift, and the charset value after a 927 ECI
- textCompaction: same 913 and 927 cases
- byteCompaction: the leading-ECI loop and the fallback-ECI read

When the marker is the last data codeword and codewords[0] equals the array length (which verifyCodewordCount allows, and the existing decode tests rely on), the read runs one past the end and throws ArrayIndexOutOfBoundsException out of PDF417Reader.decode, which is only supposed to throw ReaderException. Guarded each so a missing trailing codeword throws FormatException, same as decodeMacroBlock already does.

### DIFF
--- a/core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
+++ b/core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
@@ -115,12 +115,18 @@ final class DecodedBitStreamParser {
           codeIndex = byteCompaction(code, codewords, codeIndex, result);
           break;
         case MODE_SHIFT_TO_BYTE_COMPACTION_MODE:
+          if (codeIndex >= count) {
+            throw FormatException.getFormatInstance();
+          }
           result.append((char) codewords[codeIndex++]);
           break;
         case NUMERIC_COMPACTION_MODE_LATCH:
           codeIndex = numericCompaction(codewords, codeIndex, result);
           break;
         case ECI_CHARSET:
+          if (codeIndex >= count) {
+            throw FormatException.getFormatInstance();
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/zxing/pdf417/decoder/DecodedBitStreamParser.java
```diff
@@ -115,12 +115,18 @@ static DecoderResult decode(int[] codewords, String ecLevel) throws FormatExcept
           codeIndex = byteCompaction(code, codewords, codeIndex, result);
           break;
         case MODE_SHIFT_TO_BYTE_COMPACTION_MODE:
+          if (codeIndex >= count) {
```
srowen, line 118: I wonder if this should just be checked once, after line 108? why are these branches special? it seems like almost any branch would go on to deference codewords[codeIndex] next somehow

jmestwa-coder: only mode-shift and ECI read codewords[codeIndex] inline in this switch. the latch cases delegate to textCompaction/byteCompaction/numericCompaction, which loop on codeIndex < count themselves, and the two generic ECIs just advance the index and let the while condition catch it. a single check right after line 108 wouldn't catch this anyway, since codeIndex is still at most count there. the out-of-bounds read is the second codeword these two branches consume, so the guard has to sit on that read.

srowen (approved):

PR #2096, opened 2026-06-20
reject non-square bit matrix in qrcode BitMatrixParser

jmestwa-coder:
BitMatrixParser validates only the matrix height, so a non-square QR matrix slips through:
- the constructor checks getHeight() is >= 21 and 1 mod 4, but never that width equals the height
- readCodewords then sweeps a dimension x dimension region (unmaskBitMatrix, get(j - col, i)) indexed by the height, so it runs off a narrower matrix
- the public Decoder.decode(boolean[][]) / decode(BitMatrix) then escape with ArrayIndexOutOfBoundsException instead of the documented FormatException, e.g. new Decoder().decode(new boolean[33][8])

Reject non-square input in the constructor, the same way the Data Matrix parser already validates both dimensions.

### DIFF
--- a/core/src/main/java/com/google/zxing/qrcode/decoder/BitMatrixParser.java
+++ b/core/src/main/java/com/google/zxing/qrcode/decoder/BitMatrixParser.java
@@ -31,11 +31,11 @@ final class BitMatrixParser {
 
   /**
    * @param bitMatrix {@link BitMatrix} to parse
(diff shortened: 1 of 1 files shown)


srowen (approved):
I don't think we need a whole new test class for this, it's just overkill, but the change is OK

jmestwa-coder:
fair, dropped the standalone test class. it's just the one extra condition in the constructor now.

PR #2105, opened 2026-06-22
wrap invalid base64 as IOException in readDataURIImage

jmestwa-coder:
readDataURIImage lets Base64.getDecoder().decode throw IllegalArgumentException on a malformed data: URI, which escapes the declared IOException contract that readImage already enforces on its URL branch, so wrap the decode and rethrow it as IOException.

### DIFF
--- a/javase/src/test/java/com/google/zxing/client/j2se/ImageReaderTestCase.java
+++ b/javase/src/test/java/com/google/zxing/client/j2se/ImageReaderTestCase.java
@@ -20,6 +20,7 @@ import org.junit.Assert;
 import org.junit.Test;
 
 import java.awt.image.BufferedImage;
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


srowen (approved):

javase/src/test/java/com/google/zxing/client/j2se/ImageReaderTestCase.java
```diff
@@ -37,5 +38,10 @@ public void testDataURI() throws Exception {
     assertEquals(16, image.getWidth());
     assertEquals(16, image.getHeight());
   }
-  
+
+  @Test(expected = IOException.class)
+  public void testInvalidBase64DataURI() throws Exception {
```
srowen, line 43: I'm OK with the change, but just don't

PR #2107, opened 2026-06-22
Validate crop rectangle for TYPE_BYTE_GRAY images in BufferedImageLuminanceSource

vasiliy-mikhailov:
`BufferedImageLuminanceSource`'s constructor validates the crop rectangle (`left + width > sourceWidth || top + height > sourceHeight`) — but only in the `else` branch. For `TYPE_BYTE_GRAY` images it takes the fast path (`this.image = image`) and **skips the check entirely**, so an out-of-bounds crop on a gray image is accepted and fails later with an `ArrayIndexOutOfBoundsException` instead of the intended `IllegalArgumentException`.

This hoists the existing bounds check above the image-type branch so it applies to all image types. Adds a regression test.


srowen (approved):

srowen:
I don't know why I skipped the check initially. Probably it was 'known' that it would not be needed and this was for a tiny bit of speed on J2ME. I agree this is better.

